PR #13, opened 2019-08-02
Added support for data disk in vmss

garvct:
Added support for data disk in vmss

Will need this to build BeeGFS/Lustre with managed disks.


libexec/azhpc-build.sh
```diff
@@ -205,8 +206,24 @@ for resource_name in $(jq -r ".resources | keys | @tsv" $config_file); do
...
+            if [ "$resource_disk_count" -gt 0 ]; then
+                read_value resource_storage_sku ".resources.$resource_name.storage_sku"
+                data_cache="ReadWrite"
+                resource_disk_sizes=$(jq -r ".resources.$resource_name.data_disks | @sh" $config_file)
+                for size in $resource_disk_sizes; do
+                    if [ $size -gt 4095 ]; then
+                        data_cache="None"
+                    fi
+                done
+                data_disks_options="--storage-sku $resource_storage_sku --data-disk-sizes-gb "$resource_disk_sizes" --data-disk-caching $data_cache "
```
kanchanm: can we use the official term instead of storage sku - Storage Account Type?

garvct: I used storage sku because that's the argument that the az vmss create requires (--storage-sku)

kanchanm: ok lets keep it as such then

PR #30, opened 2019-08-29
Adding Prolb

xpillons:
Adding ProLB
[AB#7108](https://azurecat.visualstudio.com/1f98a2d2-5799-4534-825f-83ba55f0b8ab/_workitems/edit/7108)

### DIFF
--- a/apps/prolb/prolb.sh
+++ b/apps/prolb/prolb.sh
@@ -1,0 +1,166 @@
+CASE=$1
+CASE_DIR=$2
+APP_VERSION=$3
+ITER=$4
+PROFILE=$5
+
+# Set AZHPC_XXX environment variables
+AZHPC_DATA=/data
+AZHPC_APPS=/apps
+AZHPC_APPLICATION=prolb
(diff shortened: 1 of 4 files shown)


apps/prolb/prolb.sh
```diff
@@ -0,0 +1,166 @@
...
+mpi_options+=" --map-by ppr:$AZHPC_PPR:numa"
+mpi_options+=" -x LD_LIBRARY_PATH"
+mpi_options+=" -bind-to core"
+mpi_options+=" -report-bindings --display-allocation -v"
+
+if [ -n "${LD_PRELOAD}" ]; then
+    mpi_options+=" -x LD_PRELOAD"
+fi
+
+if [ "$AZHPC_VMSIZE" == "standard_hb60rs" ]; then
```
kanchanm: this may need to be updated with a new SKU so might be good to add a note in the script and the readme to update this if needed. also a note to explain how we arrived at this number for these SKUs.

xpillons: Done in the readme and the run script

PR #32, opened 2019-09-03
Add starccm+

JonShelley:
Added files and instructions to run starccm+

### DIFF
--- a/apps/starccm/readme.md
+++ b/apps/starccm/readme.md
@@ -1,0 +1,60 @@
+## Install and run Fluent Benchmarks
+
+## Prerequisites
+
+Cluster is built with the desired configuration for networking, storage, compute etc. 
+
+Recommended cluster setup
+Start with the simple_hpc_pbs example. Before you do azhpc-build copy the scripts directory from <azurehpc>/apps/starccm to your cluster build directory. First, you will need to add the following line above the "tags" section for the headnode.
+
+"data_disks": [2048, 2048],
+    
+ Second, add the following section in the scripts section above the pbsdownload piece (~line 90 in the config.json file). 
+
+{
+    "script": "add_reqs.sh",
+    "tag": "add_reqs",
+    "sudo": true
+},
+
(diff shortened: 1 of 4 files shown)


apps/starccm/readme.md
```diff
@@ -0,0 +1,65 @@
+## Install and run Fluent Benchmarks
```
kanchanm: StarCCM

apps/starccm/readme.md
```diff
@@ -0,0 +1,65 @@
...
+## Prerequisites
+
+Cluster is built with the desired configuration for networking, storage, compute etc. 
+
+Recommended cluster setup
+Start with the simple_hpc_pbs example. Before you do azhpc-build copy the scripts directory from <azurehpc>/apps/starccm to your cluster build directory. First, you will need to add the following line above the "tags" section for the headnode.
+
+"data_disks": [2048, 2048],
+    
+ Second, add the following section in the scripts section above the pbsdownload piece (~line 90 in the config.json file).
```
kanchanm: an alternative (preferred) to requiring to do this during cluster setup would be to run the script on the compute nodes post-setup. see https://github.com/Azure/azurehpc/tree/master/apps/lammps for e.g.

apps/starccm/run_civil_ompi4.pbs
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+
+#PBS -N civil
+#PBS -l select=2:ncpus=60:mpiprocs=60:mem=220gb
+#PBS -l walltime=168:0:0
+#PBS -k oed
+
+MPI=openmpi-4.0.0
+VERSION=14.06.004
+AZ_INST=HB60rs
```
kanchanm: not used..

apps/starccm/readme.md
JonShelley: Fixed

apps/starccm/readme.md
JonShelley: Addressed as recommended by the alternative approach

PR #33, opened 2019-09-05
Activedirectory

hmeiland:


scripts/join_domain.sh
```diff
@@ -0,0 +1,45 @@
...
+
+echo $1 $2 $3
+
+yum install sssd realmd oddjob oddjob-mkhomedir adcli samba-common samba-common-tools krb5-workstation openldap-clients policycoreutils-python -y
+
+systemctl restart dbus
+systemctl restart systemd-logind
+
+#NAMESERVER=$(jetpack config adjoin.ad.dns1)
+NAMESERVER=10.2.1.4
```
xpillons: The NAMESERVER is hardcoded here, shouldn't it be retrieved dynamically ?

PR #34, opened 2019-09-05
Lsf

xpillons:
Added simple LSF support
[AB#7166](https://azurecat.visualstudio.com/1f98a2d2-5799-4534-825f-83ba55f0b8ab/_workitems/edit/7166)


xpillons:
AB#7166

PR #37, opened 2019-09-09
added - Build OpenMPI with C++ bindings

xpillons:


garvct:
c++ mpi bindings were removed from the MPI 3 standard, may not be important going forward?
https://blogs.cisco.com/performance/the-mpi-c-bindings-what-happened-and-why

xpillons:
Good catch. I needed it for ProLB, which may have to update their code to avoid using these bindings.

xpillons:
> c++ mpi bindings were removed from the MPI 3 standard, may not be important going forward?
> https://blogs.cisco.com/performance/the-mpi-c-bindings-what-happened-and-why

Sure, but this was needed for ProLB.

PR #38, opened 2019-09-11
Starccm update

edwardsp:
Updated docs for running StarCCM


apps/starccm/install_starccm.sh
```diff
@@ -1,26 +1,30 @@
 #!/bin/bash
 
-sudo yum install -y unzip
+# parameters that can be overridden
+APP_INSTALL_DIR=${APP_INSTALL_DIR:-/apps}
```
xpillons: Shouldn't we start naming these variables AZHPC_XXX so they can clearly be identified as specific to this framework ?

PR #42, opened 2019-09-13
Fio single node benchmark test

garvct:
-  Script to install fio from git repo.

- PBS script to run an fio throughput and IOPS benchmark on a single Linux client.

- Powershell script to run the same fio I/O benchmark (throughput and IOPS) on a windows client.

### DIFF
--- a/apps/fio/build_fio.sh
+++ b/apps/fio/build_fio.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+APP_NAME=fio
+SHARED_APP=/apps
+MODULE_DIR=${SHARED_APP}/modulefiles
+MODULE_NAME=${APP_NAME}
+INSTALL_DIR=${SHARED_APP}/${APP_NAME}
+PARALLEL_BUILD=4
+
+function create_modulefile {
+mkdir -p ${MODULE_DIR}
+cat << EOF >> ${MODULE_DIR}/${MODULE_NAME}
+#%Module
+prepend-path PATH ${INSTALL_DIR}/bin;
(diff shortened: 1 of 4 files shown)


apps/fio/readme.md
```diff
@@ -0,0 +1,38 @@
+# FIO
+
+Fio is a general I/O benchmark code, it has many options to configure and customize the I/O pattern being tested.
+
+To build a cluster with PBS, see one of the examples, e.g. [simple_hpc_pbs](../../examples/simple_hpc_pbs/readme.md)
+
+First log-in to the headnode.
+```
```
xpillons: You should replace the connect and manual run by something like azhpc-run -u build_fio.sh.
Look at the readme.md in ProLB.

apps/fio/build_fio.sh
```diff
@@ -0,0 +1,29 @@
...
+prepend-path LD_LIBRARY_PATH ${INSTALL_DIR}/lib;
+prepend-path MAN_PATH ${INSTALL_DIR}/man;
+EOF
+}
+
+yum install -y zlib-devel
+cd $SHARED_APP
+git clone git@github.com:axboe/fio.git
+
+cd $APP_NAME
```
xpillons: Is there any prefered compiler version to use or the default one is good ?

garvct: Default compiler is good.

apps/fio/readme.md
garvct: ok, will make this change.

PR #47, opened 2019-09-16
Fix anf

hmeiland:


libexec/azhpc-build.sh
```diff
@@ -328,12 +329,15 @@ for storage_name in $(jq -r ".storage | keys | @tsv" $config_file 2>/dev/null);
...
                         --location $location \
                         --service-level $pool_service_level \
-                        --usage-threshold $(($volume_size * (2 ** 40))) \
+                        --usage-threshold $(($volume_size * (2 ** 10))) \
                         --creation-token ${volume_name} \
                         --pool-name $pool_name \
                         --volume-name $volume_name \
-                        --subnet-id $storage_subnet_id \
+ 			--vnet $vnet_name \
+                        --subnet $storage_subnet \
```
edwardsp: We need $storage_subnet_id here as we need to handle the case where the vnet is in a different resource group to ANF.  Although, I'm not convinced this was actually fixed in the azure cli - it will extract the subnet but it doesn't look as if it is doing anything with the resource group - https://github.com/Azure/azure-cli/pull/10058/files#

hmeiland:
closing and update

PR #49, opened 2019-09-17
Add anf cifs

hmeiland:
add cifs volume to anf and example config for linux client


xpillons:
@hmeiland if CLI 2.0.73 is required for running this, can you make sure that you check that version number when doing the install.sh ?

libexec/azhpc-build.sh
```diff
@@ -307,48 +331,82 @@ for storage_name in $(jq -r ".storage | keys | @tsv" $config_file 2>/dev/null);
                 mkdir -p scripts
                 status "Building script: $mount_script"
                 echo "#!/bin/bash" > $mount_script
-                echo "yum install -y nfs-utils" >> $mount_script
+                echo "yum install -y nfs-utils cifs-utils" >> $mount_script
```
edwardsp: This would be better if we only installed the packages for the protocol we want to use.

hmeiland: good point; will try to optimize...

PR #51, opened 2019-09-17
Starccm touchup

JonShelley:
Updated the installer and the readme after I hit some issues installing StarCCM

### DIFF
--- a/apps/starccm/install_starccm.sh
+++ b/apps/starccm/install_starccm.sh
@@ -3,7 +3,7 @@
 # parameters that can be overridden
 APP_INSTALL_DIR=${APP_INSTALL_DIR:-/apps}
 TMP_DIR=${TMP_DIR:-/mnt/resource}
-STARCCM_INSTALLER_FILE=${STARCCM_INSTALLER_FILE:-/mnt/resource/STAR-CCM+14.06.004_02_linux-x86_64-2.12_gnu7.1.tar.gz}
+STARCCM_INSTALLER_FILE=${STARCCM_INSTALLER_FILE:-/mnt/resource/STAR-CCM+14.04.013_01_linux-x86_64-2.12_gnu7.1.zip}
 
 if [ ! -e $STARCCM_INSTALLER_FILE ]; then
     echo "Error:  $STARCCM_INSTALLER_FILE does not exist"
--- a/apps/starccm/install_starccm.sh
+++ b/apps/starccm/install_starccm.sh
@@ -3,13 +3,11 @@
 # parameters that can be overridden
 APP_INSTALL_DIR=${APP_INSTALL_DIR:-/apps}
 TMP_DIR=${TMP_DIR:-/mnt/resource}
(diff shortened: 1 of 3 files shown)


apps/starccm/install_starccm.sh
```diff
@@ -3,13 +3,11 @@
 # parameters that can be overridden
 APP_INSTALL_DIR=${APP_INSTALL_DIR:-/apps}
 TMP_DIR=${TMP_DIR:-/mnt/resource}
-STARCCM_INSTALLER_DIR=${STARCCM_INSTALLER_DIR:-/mnt/resource}
+STARCCM_INSTALLER_FILE=${STARCCM_INSTALLER_FILE:-/mnt/resource/STAR-CCM+14.06.004_02_linux-x86_64-2.12_gnu7.1.tar.gz}
```
edwardsp: I don't think you should change the environment variable here.  We have just written a blog post using these scripts and it would stop it working.  If you would like to specify an installer you could always add another environment variable, e.g. STARCCM_ARCHIVE_NAME.  The other issue is that the run_case script is based on this installer and so would need to be updated as well for the starccm version

apps/starccm/readme.md
```diff
@@ -71,8 +76,8 @@ Environment variables can be passed to the PBS job with the `-v` flag.
 
 Submit a job as follows (remembering to substitute your PoD key value):
 
-    qsub -l select=2:ncpus=60:mpiprocs=60:place=scatter:excl \
-        -v PODKEY=#INSERT_POD_KEY# \
+    qsub -l select=2:ncpus=60:mpiprocs=60,place=scatter:excl \
+        -v PODKEY=#INSERT_POD_KEY# -v CASE=#Path to civil.sim# \
```
edwardsp: CASE is just for the name of the file excluding the path and extension.  DATA_DIR was for where the sim files are located.

apps/starccm/install_starccm.sh
kanchanm: ok with this change, i just need this in asap so i can update the blog Paul mentions :-)

apps/starccm/readme.md
kanchanm: from readme we are copying the sim file to headnode here. we need to fix that too as the script expects it in DATA_DIR as Paul mentions.
_azhpc-scp civil.sim hpcuser@headnode:._

apps/starccm/readme.md
```diff
@@ -160,8 +165,8 @@ Environment variables can be passed to the PBS job with the `-v` flag.
...
```
kanchanm: from readme we are copying the sim file to headnode here. we need to fix that too as the script expects it in DATA_DIR as Paul mentions.
_azhpc-scp civil.sim hpcuser@headnode:._

apps/starccm/install_starccm.sh
JonShelley: I think it is best to leave this change and update the blog.

PR #52, opened 2019-09-17
IO-500 HPC I/O storage benchmark

garvct:
The IO-500 I/O storage benchmark is a standard benchmark to measure and compare HPC Storage solutions.

- IO-500 installation script
- IO-500 pbs run script.

### DIFF
--- a/apps/io500/io500.pbs
+++ b/apps/io500/io500.pbs
@@ -1,0 +1,158 @@
+NSTRUCTIONS:
+# Edit this file as needed for your machine.
+# This simplified version is just for running on a single node.
+# It is a simplified version of the site-configs/sandia/startup.sh which include SLURM directives.
+# Most of the variables set in here are needed for io500_fixed.sh which gets sourced at the end of this.
+# Please also edit 'extra_description' function.
+
+FILESYSTEM=/beegfs
+
+set -euo pipefail  # better error handling
+
+io500_ior_easy_size=2000
+io500_ior_hard_writes_per_proc=30000
+
+# turn these to True successively while you debug and tune this benchmark.
+# for each one that you turn to true, go and edit the appropriate function.
+# to find the function name, see the 'main' function.
+# These are listed in the order that they run.
+io500_run_ior_easy="True" # does the write phase and enables the subsequent read
+io500_run_md_easy="True"  # does the creat phase and enables the subsequent stat
+io500_run_ior_hard="True" # does the write phase and enables the subsequent read
+io500_run_md_hard="True"  # does the creat phase and enables the subsequent read
+io500_run_find="True"
+io500_run_ior_easy_read="True"
+io500_run_md_easy_stat="True"
+io500_run_ior_hard_read="True"
+io500_run_md_hard_stat="True"
+io500_run_md_hard_read="True"
+io500_run_md_easy_delete="True" # turn this off if you want to just run find by itself
+io500_run_md_hard_delete="True" # turn this off if you want to just run find by itself
+io500_run_mdreal="False"  # this one is optional
+io500_cleanup_workdir="False"  # this flag is currently ignored. You'll need to clean up your data files manually if you want to.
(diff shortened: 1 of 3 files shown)


apps/io500/io500.pbs
```diff
@@ -0,0 +1,157 @@
+NSTRUCTIONS:
+# Edit this file as needed for your machine.
+# It is a simplified version of the site-configs/sandia/startup.sh which include SLURM directives.
+# Most of the variables set in here are needed for io500_fixed.sh which gets sourced at the end of this.
+# Please also edit 'extra_description' function.
+
+FILESYSTEM=/beegfs
```
xpillons: please consider this as a script parameter so ut can be independant from the FS to test

apps/io500/io500.pbs
```diff
@@ -0,0 +1,157 @@
+NSTRUCTIONS:
```
xpillons: Typo, missing 'I'
Can we provide a set a pre-defined parameter set based on system we already have tested ?
More generally speaking which parameters doesn't change and which one may vary ?
For those maybe add them as script params with default values so you don't need to edit the run script each time you want to change a value

apps/io500/readme.md
```diff
@@ -0,0 +1,29 @@
...
+The [IO-500](https://www.vi4io.org/std/io500/start) is a I/O storage benchmark designed to measure and compare different HPC storage solutions. It runs a suite of throughput, IOPS and metadata benchmarks and gives a score to rank the storage solution. It uses modified versions of LLNL IOR/MDtest and a find benchmark.
+
+The IO-500 benchmark needs to be built on HB or HC sku's with CentOS-HPC 7.6 (using mpi/mpich-3.3). See one of the examples for building a Cluster with HB or HC skus and PBS. (e.g. [simple_hpc_pbs](../../examples/simple_hpc_pbs/readme.md))
+
+log-on to the headnode
+```
+azhpc-connect headnode
+```
+
+Copy apps dir to the headnode
```
xpillons: You need to copy before connecting to the headnode

apps/io500/readme.md
```diff
@@ -0,0 +1,29 @@
...
+```
+azhpc-scp -r $azhpc_dir/apps hpcuser@headnode:.
+```
+
+Build the IO-500 benchmark suite from the build script.  This will install IO-500 into /apps/io-500-dev and create a modulefile:
+```
+    build_io500.sh
```
xpillons: you should replace that by a azhpc-run command. The steps would be :

- azhpc-scp
- azhpc-run
- azhpc connect
and then submit jobs

apps/io500/readme.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+Now submit and run the IO-500 benchmark (e.g on HB):
+```
+     qsub -l select=2:ncpus=60:mpiprocs=15 io500.pbs
+```
+
+> Note: this will run on 2 node and 15 processes per node.
```
xpillons: Please document when the outputs are and how they look like.

apps/io500/io500.pbs
garvct: FILESYSTEM is now passed in as a PBS variable (-v option).

apps/io500/io500.pbs
garvct: Corrected typo.
There are too many parameters to list (but I set some suitable defaults, should be able to run without editing).
Added a sample of tail of result summary file.

apps/io500/readme.md
garvct: yes, corrected that.

apps/io500/readme.md
garvct: I think its more natural to install software and run jobs from headnode (after logging into headnode).
If there is a strong preference to use azhpc-scp instead I can make that change.

apps/io500/readme.md
garvct: done.

apps/io500/readme.md
xpillons: both works, this is not a blocker, just aligned with what we are doing with other apps.

xpillons:
good for me, @garvct  you can merge it

PR #55, opened 2019-09-18
Anf full touchup

JonShelley:
The previous checkin the user had to load an az extension. These changes require the latest az cli but user no longer needs to load an az extension


garvct:
Looks good, but need to make sure there is no conflicts with PR #49 and PR #47.

PR #56, opened 2019-09-18
Starccm touchup

JonShelley:
Made some additional changes to support different versions and updated the run_case.pbs since it did not work as expected. Also updated the readme.md

### DIFF
--- a/apps/starccm/readme.md
+++ b/apps/starccm/readme.md
@@ -138,7 +138,7 @@ The following environment variables can be used:
 |-----------------------|---------------|-----------------------------------------------------------------------------------|
 | APP_INSTALL_DIR       | /apps         | The place to install (a starccm directory will be created here                    |
 | TMP_DIR               | /mnt/resource | A temporary directory for installation files                                      |
-| STARCCM_INSTALLER_DIR | /mnt/resource | The path to the `STAR-CCM+14.06.004_02_linux-x86_64-2.12_gnu7.1.tar.gz` installer |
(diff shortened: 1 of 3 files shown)


kanchanm:
@JonShelley there are some conflicts, can you merge with the latest?

apps/starccm/readme.md
```diff
@@ -17,37 +17,43 @@ azhpc-scp -r $azhpc_dir/apps hpcuser@headnode:.
...
+azhpc-run -u hpcuser -n compute apps/starccm/scripts/add_reqs.sh 
 ```
 
 ### Install Starccm+
 
 You must first obtain the starccm installer and copy it to the cluster.  If it is available on the local machine you can copy as follows:
 
 ```
-azhpc-scp STAR-CCM+14.06.004_02_linux-x86_64-2.12_gnu7.1.tar.gz /mnt/resource/.
+azhpc-scp STAR-CCM+14.04.013_01_linux-x86_64-2.12_gnu7.1.zip /mnt/resource/.
```
kanchanm: this is already fixed in the latest checkin -- should be hpcuser@headnode:/mnt/resource/.

JonShelley: added the fix.

apps/starccm/run_case.pbs
```diff
@@ -2,26 +2,29 @@
 
 # parameters that can be overridden
 APP_INSTALL_DIR=${APP_INSTALL_DIR:-/apps}
-DATA_DIR={DATA_DIR:-.}
-CASE={CASE:-civil}
-# PODKEY is required (pass in as environment variable)
-#PODKEY=
+DATA_DIR=${DATA_DIR:-/data/starccm}
+CASE=${CASE:-civil}
+# Valid OMPI values are openmpi and openmpi4
```
kanchanm: hmm.. isn't it just openmpi at this time? would suggest not mentioning this.

JonShelley: Addressed comment

PR #58, opened 2019-09-23
Segregated install

edwardsp:


edwardsp:
That is just a bug (re unable to run local post install scripts) and can be fixed in a separate commit.

PR #65, opened 2019-09-25
Added support for low pri in VMSS

edwardsp:
Add the following property in config.json

    "low_priority": true

It defaults to false.


kanchanm:
we should add a note in the readme (prerequisite) to make this more visible -- mention that this option is available, example of what to add in the config to enable it.

PR #68, opened 2019-09-25
moved pbs scripts in to the azurehpc repo

JonShelley:
Moved pbs hooks in to the azurehpc repo and updated the pbsserverhooks.sh file to reflect the change.


scripts/pbsserverhooks.sh
```diff
@@ -1,27 +1,13 @@
...
-log_analytics_account=$1
-log_analytics_key=$2
-
 cd /tmp
-sudo yum install -y git
-git clone https://github.com/JonShelley/pbs_hooks.git
-cd pbs_hooks/azure
-
-
+wget https://raw.githubusercontent.com/Azure/azurehpc/master/pbshooks/setup_jobdir.py
```
edwardsp: it would be much better if we didn't use wget here.  if you were to change the branch (or if someone cloned the repo) and made changes, this would not give the right results

scripts/pbsserverhooks.sh
```diff
@@ -1,27 +1,13 @@
...
 sudo /opt/pbs/bin/qmgr -c 'create hook setup_jobdir'
 sudo /opt/pbs/bin/qmgr -c 'set hook setup_jobdir event="execjob_begin,execjob_end"'
 sudo /opt/pbs/bin/qmgr -c 'import hook setup_jobdir application/x-python default setup_jobdir.py'
 
+wget https://raw.githubusercontent.com/Azure/azurehpc/master/pbshooks/stop_waagent.py
```
edwardsp: same as above

scripts/pbsserverhooks.sh
JonShelley: @edwardsp I have removed the wget now

scripts/pbsserverhooks.sh
JonShelley: Same as above

JonShelley:
The reason I use the job hook is because /mnt/resource on normal VMs is not user readable. I believe that we make /mnt/resource world writable in azhpc so maybe it is not needed. But on the other hand it creates a dir with the job id and the user permissions on mount resource if they need the local resource.

JonShelley:
Where do we currently put documentation for our scripts? Should I create a readme.md file in the pbshooks dir?

edwardsp:
I think a readme in the pbshooks dir is good

JonShelley:
> If you were to move the pbshooks into the scripts directory you would already have them on all the nodes as **all** scripts get copied. This would solve the wget issue raised below.
> 
> We need to start adding some documentation for this as well. I understand stopping waagent but I'm not sure why you would want the setup_jobdir hook.

@edwardsp I have added a readme. Are you satisfied with the commit and ready to approve it or do you have other concerns?

PR #69, opened 2019-09-26
Added manageddisk_lustre for Lustre managed disks in the OSS and MDS …

chadnar2:
Additions of managed disks to Lustre MDS and OSS servers.

### DIFF
--- a/examples/manageddisk_lustre/config.json
+++ b/examples/manageddisk_lustre/config.json
@@ -156,7 +156,7 @@
       "script": "create_raid0.sh",
       "args": [
         "/dev/md0",
-        "/dev/sdc /dev/sdd /dev/sde /dev/sdf /dev/sdg /dev/sdh"
+	"/dev/sd[c-l]"
       ],
       "tag": "lfsoss",
       "sudo": true
--- a/examples/manageddisk_lustre/config.json
+++ b/examples/manageddisk_lustre/config.json
@@ -1,0 +1,257 @@
+{
+  "location": "westeurope",
+  "resource_group": "variables.resource_group",
+  "install_from": "headnode",
+  "admin_user": "hpcadmin",
+  "vnet": {
+    "name": "hpcvnet",
+    "address_prefix": "10.2.0.0/20",
+    "subnets": {
+      "compute": "10.2.0.0/22",
+      "storage": "10.2.4.0/24"
+    }
+  },
+  "variables": {
+    "resource_group": "azhpc-cluster",
+    "lustre_disk_type": "data_disk",
+    "image": "OpenLogic:CentOS:7.6:latest",
+    "hpcimage": "OpenLogic:CentOS-HPC:7.6:latest",
+    "storage_account": "xxxxxxxx",
+    "storage_key": "/xxxxxxxxxxxxx",
+    "storage_container": "xxxxxxxxx",
+    "log_analytics_lfs_name": "xxxxxxx",
+    "log_analytics_workspace": "xxxxxxxx",
+    "log_analytics_key": "xxxxxxx",
+    "location": "westeurope"
+  },
+  "resources": {
+    "headnode": {
+      "type": "vm",
+      "vm_type": "Standard_HB60rs",
+      "accelerated_networking": false,
+      "public_ip": true,
+      "image": "variables.hpcimage",
+      "subnet": "compute",
+      "tags": [
+        "disable-selinux",
+        "lfsclient",
+        "lfsazimport",
+        "localuser",
+        "pbsserver",
+        "loginnode",
+        "nfsserver"
+      ]
+    },
+    "compute": {
+      "type": "vmss",
+      "vm_type": "Standard_HB60rs",
+      "accelerated_networking": false,
+      "instances": 2,
+      "image": "variables.hpcimage",
+      "subnet": "compute",
+      "tags": [
+        "disable-selinux",
+        "lfsclient",
+        "localuser",
+        "pbsclient",
+        "nfsclient"
+      ]
+    },
+    "lfsmds": {
+      "type": "vm",
+      "vm_type": "Standard_D16s_v3",
+      "accelerated_networking": true,
+      "image": "variables.image",
+      "subnet": "storage",
+      "storage_sku": "Premium_LRS",
+      "data_disks": [
+        4096
+      ],
+      "tags": [
+        "lfspkgs",
+        "lfsmaster",
+        "disable-selinux",
+        "lfsloganalytics"
+      ]
+    },
+    "lfsoss": {
+      "type": "vmss",
+      "vm_type": "Standard_D16s_v3",
+      "instances": 2,
+      "accelerated_networking": true,
+      "image": "variables.image",
+      "subnet": "storage",
+      "storage_sku": "Premium_LRS",
+      "data_disks": [
+        1024,
+        1024,
+        1024,
+        1024,
+        1024,
+        1024
+      ],
+      "tags": [
+        "lfspkgs",
+        "lfsoss",
+        "disable-selinux",
+        "lfsloganalytics"
+      ]
+    },
+    "lfshsm": {
+      "type": "vmss",
+      "vm_type": "Standard_D16s_v3",
+      "instances": 1,
+      "accelerated_networking": true,
+      "image": "variables.image",
+      "subnet": "storage",
+      "tags": [
+        "lfspkgs",
+        "lfshsm",
+        "disable-selinux"
+      ]
+    },
+    "lfssmb": {
+      "type": "vm",
+      "vm_type": "Standard_D16s_v3",
+      "instances": 1,
+      "accelerated_networking": true,
+      "image": "variables.image",
+      "subnet": "storage",
+      "tags": [
+        "lfspkgs",
+        "lfsclient",
+        "lfssmb",
+        "disable-selinux",
+        "localuser"
+      ]
+    }
+  },
+  "install": [
+    {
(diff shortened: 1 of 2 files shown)


examples/manageddisk_lustre/config.json
```diff
@@ -0,0 +1,257 @@
...
+      "compute": "10.2.0.0/22",
+      "storage": "10.2.4.0/24"
+    }
+  },
+  "variables": {
+    "resource_group": "azhpc-cluster",
+    "lustre_disk_type": "data_disk",
+    "image": "OpenLogic:CentOS:7.6:latest",
+    "hpcimage": "OpenLogic:CentOS-HPC:7.6:latest",
+    "storage_account": "xxxxxxxx",
```
xpillons: Please use <NET-SET> instead of xxxx as these will be automatically parsed by the azhpc-init

examples/manageddisk_lustre/config.json
```diff
@@ -0,0 +1,257 @@
...
+    "storage_container": "xxxxxxxxx",
+    "log_analytics_lfs_name": "xxxxxxx",
+    "log_analytics_workspace": "xxxxxxxx",
+    "log_analytics_key": "xxxxxxx",
+    "location": "westeurope"
+  },
+  "resources": {
+    "headnode": {
+      "type": "vm",
+      "vm_type": "Standard_HB60rs",
```
xpillons: Please don't use HB for Headnode for cost savings

examples/manageddisk_lustre/config.json
```diff
@@ -0,0 +1,257 @@
...
+  "vnet": {
+    "name": "hpcvnet",
+    "address_prefix": "10.2.0.0/20",
+    "subnets": {
+      "compute": "10.2.0.0/22",
+      "storage": "10.2.4.0/24"
+    }
+  },
+  "variables": {
+    "resource_group": "azhpc-cluster",
```
edwardsp: put this to <NOT-SET> as the script checks for this

examples/manageddisk_lustre/config.json
```diff
@@ -0,0 +1,257 @@
...
+        "lfsazimport",
+        "localuser",
+        "pbsserver",
+        "loginnode",
+        "nfsserver"
+      ]
+    },
+    "compute": {
+      "type": "vmss",
+      "vm_type": "Standard_HB60rs",
```
xpillons: Do you really need HB as an example for mounting Lustre ?

examples/manageddisk_lustre/config.json
```diff
@@ -0,0 +1,257 @@
...
+    "name": "hpcvnet",
+    "address_prefix": "10.2.0.0/20",
+    "subnets": {
+      "compute": "10.2.0.0/22",
+      "storage": "10.2.4.0/24"
+    }
+  },
+  "variables": {
+    "resource_group": "azhpc-cluster",
+    "lustre_disk_type": "data_disk",
```
edwardsp: this is unused

examples/manageddisk_lustre/config.json
```diff
@@ -0,0 +1,257 @@
...
+    "image": "OpenLogic:CentOS:7.6:latest",
+    "hpcimage": "OpenLogic:CentOS-HPC:7.6:latest",
+    "storage_account": "xxxxxxxx",
+    "storage_key": "/xxxxxxxxxxxxx",
+    "storage_container": "xxxxxxxxx",
+    "log_analytics_lfs_name": "xxxxxxx",
+    "log_analytics_workspace": "xxxxxxxx",
+    "log_analytics_key": "xxxxxxx",
+    "location": "westeurope"
```
edwardsp: you have a location variable but it is hard coded earlier in the config file

examples/manageddisk_lustre/config.json
```diff
@@ -0,0 +1,257 @@
...
+        "$(<hostlists/tags/nfsserver)"
+      ],
+      "tag": "localuser",
+      "sudo": true
+    },
+    {
+      "script": "create_raid0.sh",
+      "args": [
+        "/dev/md0",
+        "/dev/sdc /dev/sdd /dev/sde /dev/sdf /dev/sdg /dev/sdh"
```
xpillons: can you discover these dynamically instead of passing them as an arg ? please look at the nfsserver.sh script which contains that code portion

examples/manageddisk_lustre/config.json
```diff
@@ -0,0 +1,257 @@
+{
+  "location": "westeurope",
```
edwardsp: this is a hard coded value when there is a variable for it defined below

examples/manageddisk_lustre/config.json
edwardsp: use <NOT-SET>

examples/manageddisk_lustre/config.json
edwardsp: you could try simplifying this is bash expansion by doing something like:

    "/dev/sd[c-l]"

That *should* pick up the first 10 disks (assuming it starts with sdc and is consecutive) so you can change the disks in the config file without needing to change the arguments here.

You would need to test this though.

examples/manageddisk_lustre/config.json
chadnar2: ok

examples/manageddisk_lustre/config.json
chadnar2: But this is what was done for all the Lustre versions for azhpc, not just here. I need a clear determination for the compute nodes.

examples/manageddisk_lustre/config.json
chadnar2: This is what was done for Lustre previously with Azure. Either we change all of them or none of them I would think.

examples/manageddisk_lustre/config.json
chadnar2: ok

examples/manageddisk_lustre/config.json
chadnar2: I saw the previous lustre_full also has this coded in the config file. I did make the change to "<NOT-SET>"

examples/manageddisk_lustre/config.json
chadnar2: Changed to "<NOT-SET>", but this should be done with lustre_full too then.

examples/manageddisk_lustre/config.json
chadnar2: done

examples/manageddisk_lustre/config.json
chadnar2: Something will be changed, maybe even auto-detecting the new drives.

chadnar2:
I have some concerns with these types of comments showing up during the azhpc-connects 
root@MININT-JHIRRML:/manageddisk_lustre2# azhpc-connect lfsosd34f000003
Thu Sep 26 12:02:40 CDT 2019 : logging in to lfsosd34f000003 (via headnode06036d.westeurope.cloudapp.azure.com)
The Resource 'Microsoft.Compute/virtualMachines/lfsosd34f000003' under resource group 'narazhpc-cluster' was not found.
[hpcadmin@lfsosd34f000003 ~]$

but it was found, but maybe not under Compute????

examples/manageddisk_lustre/config.json
edwardsp: it's fine to put a location here in the variables.  WHat you need to do is then reference it in the config file.  At the moment, nothing uses this variable, it is hardcoded as "westeurope" at the top.

examples/manageddisk_lustre/config.json
edwardsp: What do you mean here?  My above suggestion should find up to 10 premium disks that are added (although it needs testing).  If they are not attached, they should not appear in the list.

examples/manageddisk_lustre/config.json
edwardsp: lustre_full has a hardcoded location at the top but it doesn't have an unused location variable.  Having an variable would confuse someone as they would expect it to actually change the location.

examples/manageddisk_lustre/config.json
chadnar2: changed

examples/manageddisk_lustre/config.json
chadnar2: changed

examples/manageddisk_lustre/config.json
xpillons: Please start by changing this one, we will change the others too.

examples/manageddisk_lustre/config.json
xpillons: Please start by changing this one, we will change the others too.

examples/manageddisk_lustre/config.json
chadnar2: I thought the decision was to use HB nodes.

xpillons: No because this is the most expensive choice.

examples/manageddisk_lustre/config.json
chadnar2: Change made. Code ready for merge, but please discuss Compute Node type with Xavier. Currently using HB servers.

chadnar2: Up to 10 drives will be automatically added now per OSS.

examples/manageddisk_lustre/config.json
chadnar2: Please discuss with Paul. Changes can be made directly to the config.json as determined.

PR #72, opened 2019-09-27
Health checks

JonShelley:
Updated OSU BW test to run in parallel with PBS, added stream test, and updated readme to explain how to run it.


xpillons:
@JonShelley we discussed this already with @garvct and @kanchanm on how to organize these heathchecks. Please move the OSU script in to the /apps/osu directory the healthchecks should run these and do checkings of the returned value based on the SKU size.

JonShelley:
> @JonShelley we discussed this already with @garvct and @kanchanm on how to organize these heathchecks. Please move the OSU script in to the /apps/osu directory the healthchecks should run these and do checkings of the returned value based on the SKU size.

I'm not sure I understand. The changes I made were to run the ring test in parallel using a HPC image and PBS. This is a health check in my mind and is in the right location. Am I missing something? @kanchanm  and @garvct , thoughts?

PR #73, opened 2019-09-30
Bubble up errors when running scripts

xpillons:
Bubble up script errors when running either locally or remotely. The default behavior is to failed in case of any script returning exit code !=0

### DIFF
--- a/libexec/install_helper.sh
+++ b/libexec/install_helper.sh
@@ -240,6 +240,7 @@ function run_install_scripts()
         fi
     done
 
+    script_error=0
     for step in $(seq 0 $nsteps); do
 
         # skip jumpbox setup if no jumpbox scripts are required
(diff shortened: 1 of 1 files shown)


libexec/install_helper.sh
```diff
@@ -271,34 +272,47 @@ function run_install_scripts()
...
+                script_error=1
+                break
+            fi
 
         elif [ "$install_script_type" = "local_script" ]; then
-            
+
             $install_sh
+            if [ "$?" -ne "0" ]; then
+                echo "Error: Errors while running $install_sh"
```
edwardsp: Can we add the exit code here?

libexec/install_helper.sh
```diff
@@ -271,34 +272,47 @@ function run_install_scripts()
...
 
         echo "Step $step : $install_script ($install_script_type)"
         start_time=$SECONDS
 
         if [ "$install_script_type" = "jumpbox_script" ]; then
-        
+
             ssh $ssh_args -i $ssh_private_key $admin_user@$fqdn $install_sh $run_tag
+            if [ "$?" -ne "0" ]; then
+                echo "Error: Errors while running $install_sh"
```
edwardsp: The exit code would be good

PR #77, opened 2019-10-02
allow variable references in key,fqdn,sakey,sasurl values

xpillons:
Provide the capability to reference variables inside other values so you can set a key_vault name as a variables instead of hardcoded it in the secret.<key-vault>.<secret> value.


xpillons:
no longer needed

PR #88, opened 2019-10-15
Blobfuse

hmeiland:


scripts/blobfuse.sh
```diff
@@ -0,0 +1,20 @@
...
+
+export AZURE_STORAGE_ACCOUNT=${blob_account}
+export AZURE_STORAGE_ACCESS_KEY="${blob_key}"
+
+sudo mkdir -p /mnt/resource/blobfusetmp
+sudo chmod 777 /mnt/resource/blobfusetmp
+sudo mkdir /${blob_mount}
+sudo chmod 777 /${blob_mount}
+
+blobfuse /${blob_mount} --container-name=${blob_container} --tmp-path=/mnt/resource/blobfusetmp
```
xpillons: What if the node is rebooted ? this is going to failed as rge blobfusetmp dir no longer exists

examples/blobfuse/config.json
```diff
@@ -0,0 +1,72 @@
...
+    "name": "hpcvnet",
+    "address_prefix": "10.2.0.0/20",
+    "subnets": {
+      "compute": "10.2.4.0/22"
+    }
+  },
+  "resources": {
+    "headnode": {
+      "type": "vm",
+      "vm_type": "Standard_HC44rs",
```
xpillons: Please stop using HC when not needed. Use Standard_D8s_v3 which is enough for this kind of testing

PR #95, opened 2019-10-25
Js pipelines

JonShelley:
Added a variable for the compute instances so that we could easily update them in a pipeline


chadnar2:
Approval done

PR #103, opened 2019-11-15
Support proximity placement groups

garvct:
-Support proximity placement groups (ppg) for resource types (vn and vmss).

This will allow IaaS resources to be deployed in the same datacenter (reducing latency). e.g deploy luster/beegfs and compute nodes in the same datacenter (Best practices).

### DIFF
--- a/libexec/azhpc-build.sh
+++ b/libexec/azhpc-build.sh
@@ -216,7 +216,7 @@ for resource_name in $(jq -r ".resources | keys | @tsv" $config_file); do
             read_value resource_vm_type ".resources.$resource_name.vm_type"
             read_value resource_image ".resources.$resource_name.image"
             read_value resource_pip ".resources.$resource_name.public_ip" false
-            read_value ppg ".resources.$resource_name.proximity_placement_group" false
+            read_value resource_ppg ".resources.$resource_name.proximity_placement_group" false
             read_value resource_subnet ".resources.$resource_name.subnet"
             read_value resource_an ".resources.$resource_name.accelerated_networking" false
             resource_disk_count=$(jq -r ".resources.$resource_name.data_disks | length" $config_file)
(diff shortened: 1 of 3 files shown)


libexec/azhpc-build.sh
```diff
@@ -63,6 +63,9 @@ read_value vnet_name ".vnet.name"
 read_value address_prefix ".vnet.address_prefix"
 read_value admin_user ".admin_user"
 read_value install_node ".install_from"
+if [ ! $(jq -r ".ppg_name" $config_file) = "null" ]; then
```
edwardsp: you can just read with a default value here instead of checking with jq

garvct: I removed the if check and set a default instead. I also change ppg to be more description (proximity_placement_group).

libexec/azhpc-build.sh
```diff
@@ -303,6 +321,10 @@ for resource_name in $(jq -r ".resources | keys | @tsv" $config_file); do
             if [ "$resource_lowpri" = "true" ]; then
                 lowpri_option="--priority Low"
             fi
+            ppg_option=
+            if [ "$ppg" = "true" ]; then
+               ppg_option="--ppg $ppg_name"
```
edwardsp: this will fail here if there is no proximity group name - should we check and give an error?

libexec/azhpc-build.sh
```diff
@@ -205,6 +216,7 @@ for resource_name in $(jq -r ".resources | keys | @tsv" $config_file); do
             read_value resource_vm_type ".resources.$resource_name.vm_type"
             read_value resource_image ".resources.$resource_name.image"
             read_value resource_pip ".resources.$resource_name.public_ip" false
+            read_value ppg ".resources.$resource_name.proximity_placement_group" false
```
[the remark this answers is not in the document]
garvct: I changed the variable ppg to be resource_ppg.

libexec/azhpc-build.sh
```diff
@@ -274,10 +290,12 @@ for resource_name in $(jq -r ".resources | keys | @tsv" $config_file); do
             read_value resource_subnet ".resources.$resource_name.subnet"
             read_value resource_an ".resources.$resource_name.accelerated_networking" false
             read_value resource_lowpri ".resources.$resource_name.low_priority" false
+            read_value ppg ".resources.$resource_name.proximity_placement_group" false
```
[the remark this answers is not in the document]
garvct: Corrected.

libexec/azhpc-build.sh
garvct: Added an error check.

PR #105, opened 2019-11-29
updates for LSF and CentOS 7.7

xpillons:


apps/ior/build_ior.sh
```diff
@@ -10,7 +10,8 @@ IOR_VERSION=3.2.1
 sudo yum install -y jq
 
 source /etc/profile.d/modules.sh # so we can load modules
-module load gcc-8.2.0
+# GCC 8 is no longer provided with CentOS-HPC 7.7 image, it is now 9.2, but is this really needed ?
```
garvct: Does build_ior.sh also need to be modified (i.e its built with 8.2.0)?
If IOR is built with GCC 8.2.0 or 9.2, but at runtime GCC is not loaded is there any issues at runtime (missing shared library or incorrect GCC version symbol?)

PR #109, opened 2019-12-03
Spack installation (and accessing buildcache from blob storage)

garvct:
-spack installation
-PBS scripts, examples running osu-micro-benchmarks using openmpi, hpcx, mvapich2 and intel mpi (CentOS-HPC 7.7)
-Detailed documentation


garvct:
Thanks Paul, Valid points, I created separate configuration files and simplified the spack install script.

PR #110, opened 2019-12-03
Updated fluent scripts

JonShelley:
Updated the fluent install and run script to use the latest version


apps/fluent/run_fluent_hpcx.sh
```diff
@@ -68,8 +70,8 @@ fluentbench.pl \
     $MODEL \
     -t$CORES \
     -pinfiniband \
-    -mpi=openmpi \
-    -mpiopt="-mca btl ^vader,tcp,openib --mca opal_warn_on_missing_libcuda 0 -mca plm_rsh_no_tree_spawn 1 -mca plm_rsh_num_concurrent 300 -mca plm_base_verbose 5 -mca routed_base_verbose 5 -bind-to core -map-by ppr:$ppr:numa -report-bindings -x UCX_NET_DEVICES=mlx5_0:1 -x UCX_IB_PKEY=$PKEY -mca btl_openib_if_include mlx5_0:1 -x UCX_TLS=ud,sm,self" \
+    -mpi=$OMPI \
+    -mpiopt="--mca opal_warn_on_missing_libcuda 0 -mca plm_rsh_no_tree_spawn 1 -mca plm_rsh_num_concurrent 300 -bind-to core -map-by node -report-bindings " \
```
edwardsp: Why have we removed binding to NUMA domains?  Running 60 ppn will be on a single socket with this update.

JonShelley: I was trying to clean it up so that we did not have to run exact divisions of cores to match numa nodes. If a user wants to do this they can but on a HB the best runs were at 56 ppn and it would not work with this setting.

PR #111, opened 2019-12-03
Added vmss option for platform fault domain count

JonShelley:
On the HBv2 systems when creating a VMSS over ~94 nodes you need to specify the fault domain count or the vmss will not create. I have added the option so that it can be used in a config file if needed


libexec/azhpc-build.sh
```diff
@@ -274,6 +274,7 @@ for resource_name in $(jq -r ".resources | keys | @tsv" $config_file); do
             read_value resource_vm_type ".resources.$resource_name.vm_type"
             read_value resource_image ".resources.$resource_name.image"
             read_value resource_subnet ".resources.$resource_name.subnet"
+            read_value resource_fault_domain_count ".resources.$resource_name.fault_domain_count" 5
```
garvct: Why is the default 5? Will this change have any impact on HBv1 and HC?

JonShelley: The current default in the CLI is 5. So I am making no change to the default.

PR #112, opened 2019-12-05
Added macros for log analytics

edwardsp:


libexec/common.sh
```diff
@@ -107,6 +107,28 @@ function process_value {
         debug "getting storage key for $storage_name"
         local storage_key=$(az storage account keys list -n $storage_name --query "[0].value" | sed 's/\"//g')
         read $1 <<< "$storage_key"
+    elif [ "$prefix" = "laworkspace" ]; then
```
xpillons: Finally thought on the syntax can't it be la.workspace and la.key. This doesn't change a lot but more organized by object type

edwardsp: it's a good idea - but we probably need to update existing macros for this format and it'll be a breaking change - I'll leave for now

PR #114, opened 2019-12-12
Private dns

hmeiland:
add private dns functionality for reversed lookups...


azurehpc-ui/src/AzurehpcApp.js
```diff
@@ -158,4 +158,4 @@ class AzurehpcApp extends React.Component {
     }
 }
 
-export default AzurehpcApp;
\ No newline at end of file
+export default AzurehpcApp;
```
xpillons: This file shouldn't be in that PR

Issue #119, opened 2019-12-16
BeeGFS client issue


lmiroslaw:
After deploying https://github.com/Azure/azurehpc/tree/master/examples/beegfs

I get: 
Stderr: Job for beegfs-client.service failed because the control process exited with error code. See "systemctl status beegfs-client.service" and "journalctl -xe" for details

CC team had a similar issue with rdma drivers being not available for the client to converge.

garvct:
Can you try CentOS-HPC 7.7 and see if you see the same error?

lmiroslaw:
the error is the same with CentOS-HPC 7.7:

Stderr: Job for beegfs-client.service failed because the control process exited with error code. See "systemctl status beegfs-client.service" and "journalctl -xe" for details.

BTW, the same problem is for beegfs-data-disk

garvct:
Thanks, for checking.

It looks like the beegfs packages were updated on Nov 29, 2019 from 7.1.3 to 7.1.4. It appears that this update broke BeeGFS deployed in azurehpc on CentOS 7.7. I have created a branch called "beegfs_713" which deploys BeeGFS using the 7.1.3 packages instead of 7.1.4, all BeeGFS examples have been updated in this branch. I will create a PR later. Please try branch beegfs_713 and see if you can now deploy correctly.

garvct:
Added PR #120

lmiroslaw:
it worked! I just tested it with beegfs example. Thanks @garvct.

PR #120, opened 2019-12-16
Beegfs error correction

garvct:
-BeeGFS update to 7.1.4 does not deploy correctly.
-Add BeeGFS 7.1.3 version
-Added 7.1.3 version of beegfspkgs-713.sh
-Updated BeeGFS examples
-issue #119


scripts/beegfspkgs-713.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
xpillons: How much differences differences exists in the scripts between versions ?
Can't you just use the same script and have the version number as a parameter with a default value to keep existing config file untouched. This will make easier the update for future releases

garvct: yes, good point. I am hoping this will be a temporary fix and that ba ug fix will be released from thinkparQ and we can eventually remove this special version.

garvct: I removed beegfspkgs-713.sh and added changed to beegfspkgs.sh (so beegfs examples do not need to be modified).

scripts/beegfspkgs.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 #
+BEEGFS_VER=713
```
xpillons: You may change this to ${1-713} so that you can pass it as a parameter with a default value to 713

scripts/beegfspkgs.sh
```diff
@@ -11,7 +13,13 @@ yum install -y kernel-devel-${KERNEL} kernel-headers-${KERNEL} kernel-tools-libs
 yum -y install gcc gcc-c++
 yum -y install zlib zlib-devel bzip2 bzip2-devel bzip2-libs openssl openssl-devel openssl-libs nfs-utils rpcbind mdadm wget python-pip automake autoconf
 #
+if [ ${BEEGFS_VER} -eq 713 ]
```
xpillons: You may test if BEEGFS_VER contains a value and in that case use that variable in the wget and the sed so that your script will work for any other versions

garvct:
Abandon this branch and use beegfs_fx instead.

PR #127, opened 2019-12-23
Corrected wps installation

garvct:
- Added packages (jasper-devel & libpng-devel)
-Corrected path to WRF installation.

### DIFF
--- a/apps/wrf/install_wps_mvapich2.sh
+++ b/apps/wrf/install_wps_mvapich2.sh
@@ -7,6 +7,9 @@ SHARED_APP=/apps
 WRF_VERSION=4.1.3
 APP_DIR=${SHARED_APP}/${SKU_TYPE}/${APP_NAME}-mvapich2
 
+sudo yum install -y jasper-devel
+sudo yum install -y libpng-devel
+
 mkdir -p ${APP_DIR}
 cd ${APP_DIR}
 wget https://github.com/wrf-model/WPS/archive/v${APP_VERSION}.tar.gz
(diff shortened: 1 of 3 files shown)


apps/wrf/install_wps_mvapich2.sh
```diff
@@ -23,7 +26,7 @@ module load gcc-9.2.0
 
 export HDF5=$(spack location -i hdf5^mvapich2)
 export NETCDF=$(spack location -i netcdf-fortran^mvapich2)
-export WRF_DIR=${SHARED_APP}/WRF-${WRF_VERSION}
+export WRF_DIR=${SHARED_APP}/${SKU_TYPE}/wrf-mvapich2/WRF-${WRF_VERSION}
```
xpillons: Can't the APP_DIR be reused here ?
Is there any differences between WRF_VERSION and APP_VERSION ?

apps/wrf/install_wps_openmpi.sh
```diff
@@ -23,7 +26,7 @@ module load gcc-9.2.0
 
 export HDF5=$(spack location -i hdf5^openmpi)
 export NETCDF=$(spack location -i netcdf-fortran^openmpi)
-export WRF_DIR=${SHARED_APP}/WRF-${WRF_VERSION}
+export WRF_DIR=${SHARED_APP}/${SKU_TYPE}/wrf-openmp/WRF-${WRF_VERSION}
```
xpillons: is it wrf_openmpi or wrf-openmpi ?
Same feedbacks as for mvapich2 regarding reusing APP_DIR

apps/wrf/install_wps_mvapich2.sh
garvct: The WPS installation needs the path to WRF and so APP_DIR cannot be used (it is location of WPS and not WRF).
APP_VERSION is WPS version (4.1), but I also need WRF version (WRF_VERSION) which is different, 4.1.3.

apps/wrf/install_wps_openmpi.sh
garvct: wrf-openmpi is correct (not wrf_openmpi).

PR #128, opened 2019-12-28
WRF update

garvct:
-Loaded mpi libraries with module instead of spack.
-Added hybrid parallel builds (openmp + mpi) of wrf.

### DIFF
--- a/apps/wrf/install_wrf_omp_mvapich2.sh
+++ b/apps/wrf/install_wrf_omp_mvapich2.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+APP_NAME=wrf
+APP_VERSION=4.1.3
+SKU_TYPE=hb
+SHARED_APP=/apps
+MODULE_DIR=${SHARED_APP}/modulefiles/${SKU_TYPE}/${APP_NAME}
+MODULE_NAME=${APP_VERSION}-omp-mvapich2
+APP_DIR=$SHARED_APP/${SKU_TYPE}/${APP_NAME}-omp-mvapich2
+APPS_WRF_DIR=`pwd`
+
+function create_modulefile {
+mkdir -p ${MODULE_DIR}
+cat << EOF >> ${MODULE_DIR}/${MODULE_NAME}
+#%Module
+set              wrfversion        ${APP_VERSION}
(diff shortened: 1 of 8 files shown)


apps/wrf/install_wrf_omp_mvapich2.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+APP_NAME=wrf
+APP_VERSION=4.1.3
+SKU_TYPE=hb
```
JonShelley: I would recommend that we change this to SKU_TYPE==${SKU_TYPE:-hb}

apps/wrf/install_wrf_omp_mvapich2.sh
```diff
@@ -0,0 +1,60 @@
...
+SHARED_APP=/apps
```
JonShelley: I would recommend that this be changed to SHARED_APP=${SHARED_APP:-/apps}

apps/wrf/install_wrf_omp_mvapich2.sh
```diff
@@ -0,0 +1,60 @@
...
+MODULE_DIR=${SHARED_APP}/modulefiles/${SKU_TYPE}/${APP_NAME}
```
JonShelley: I recommend that we make this consistent with our other installers that allow you to set environment variables before running the installer so that the user does not have to change the install scripts if they want to run on different SKUs or place the application in a different directory

apps/wrf/install_wrf_omp_openmpi.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+APP_NAME=wrf
+APP_VERSION=4.1.3
+SKU_TYPE=hb
```
JonShelley: I would recommend that this be changed to SKU_TYPE=${SKU_TYPE:-hb}

apps/wrf/install_wrf_omp_openmpi.sh
```diff
@@ -0,0 +1,60 @@
...
+SHARED_APP=/apps
```
JonShelley: I would recommend that this be changed to SHARED_APP=${SHARED_APP:-/apps}

PR #134, opened 2020-01-10
Storage updates

edwardsp:
adding options for os disk size and storage types


libexec/azhpc-build.sh
```diff
@@ -331,16 +345,24 @@ for resource_name in $(jq -r ".resources | keys | @tsv" $config_file); do
             read_value resource_subnet ".resources.$resource_name.subnet"
             read_value resource_fault_domain_count ".resources.$resource_name.fault_domain_count" 5
             read_value resource_an ".resources.$resource_name.accelerated_networking" false
+            read_value resource_storage_sku ".resources.$resource_name.storage_sku" StandardSSD_LRS
+            read_value resource_os_disk ".resources.$resource_name.os_disk" 32
```
garvct: Might consider changing os_disk to os_disk_size (to be clearer).

edwardsp: good point - updated.

PR #139, opened 2020-01-15
bux fix to run local scripts without a jumpbox

edwardsp:


hmeiland:
seems to work good now for local_scripts

PR #143, opened 2020-01-20
Switched to using spot

edwardsp:
Added for both VMs and VMSSs


libexec/azhpc-build.sh
```diff
@@ -259,6 +259,7 @@ for resource_name in $(jq -r ".resources | keys | @tsv" $config_file); do
             read_value resource_subnet ".resources.$resource_name.subnet"
             read_value resource_an ".resources.$resource_name.accelerated_networking" false
             read_value resource_storage_sku ".resources.$resource_name.storage_sku" StandardSSD_LRS
+            read_value resource_lowpri ".resources.$resource_name.low_priority" false
```
xpillons: Maybe you should introduce a new param for spot as well

PR #144, opened 2020-01-22
Added AllReduce for IMPI and OMPI

xpillons:
Added AllReduce test with IMPI and OMPI
Added telemetry extraction
Added IMPI options for pingpong test


apps/imb-mpi/ringpingpong.pbs
```diff
@@ -2,24 +2,31 @@
 
 # setup Intel MPI environment for Infiniband
 source /etc/profile # so we can load modules
-module load mpi/impi_2018.4.274
+module load mpi/impi
```
edwardsp: why are we dropping the version here?

apps/imb-mpi/ringpingpong.pbs
```diff
@@ -2,24 +2,31 @@
...
 source $MPI_BIN/mpivars.sh
 
 hostlist=$(pwd)/hosts.$PBS_JOBID
 
 sort -u $PBS_NODEFILE > $hostlist
 
+export I_MPI_FABRICS="shm:dapl"
+export I_MPI_DYNAMIC_CONNECTION=0
```
edwardsp: this is only needed for h16r (and due to the lack of support from the microsoft driver)

apps/imb-mpi/ringpingpong.pbs
```diff
@@ -2,24 +2,31 @@
...
 source /etc/profile # so we can load modules
-module load mpi/impi_2018.4.274
+module load mpi/impi
 source $MPI_BIN/mpivars.sh
 
 hostlist=$(pwd)/hosts.$PBS_JOBID
 
 sort -u $PBS_NODEFILE > $hostlist
 
+export I_MPI_FABRICS="shm:dapl"
```
edwardsp: this won't work with later versions of intel mpi (and you dropped the version from the module load)

apps/imb-mpi/ringpingpong.pbs
xpillons: impi is now the generic one for 2018

apps/imb-mpi/ringpingpong.pbs
xpillons: got it

apps/imb-mpi/ringpingpong.pbs
xpillons: yes I need to change this based on IMPI version

PR #157, opened 2020-02-04
namd application

garvct:
-Initial version of namd 2.13 (build from source code).
-Supports building smp version
-Experimental memopt version support.
-Contains PBS run script.

### DIFF
--- a/apps/namd/install_namd_openmpi.sh
+++ b/apps/namd/install_namd_openmpi.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+#
+# Relevant environmental variables
+#
+# SKU_TYPE = hb, hc or hbv2 (default hbv2)
+# NAMD_SOURCE_TAR_GZ_LOC, location of NAMD_2.13_Source.tar.gz source code (default hpcuser home dir)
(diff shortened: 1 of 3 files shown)


apps/namd/install_namd_openmpi.sh
```diff
@@ -0,0 +1,93 @@
...
+#
+# SKU_TYPE = hb, hc or hbv2 (default hbv2)
+# NAMD_SOURCE_TAR_GZ_LOC, location of NAMD_2.13_Source.tar.gz source code (default hpcuser home dir)
+# NAMD_SMP, set to "smp" if you want to build the smp version (default build non smp version)
+# NAMD_MEMOPT, set to "--with-memopt" if you want to build memopt enabled version (default : memopt is disabled) 
+#
+#
+APP_NAME=NAMD
+APP_VERSION=2.13
+SHARED_APP=/avere/apps
```
xpillons: Can you make this an input variable as root is not going to be avere for all installations

garvct: yes, thanks for spotting this error. I removed the avere reference.

PR #165, opened 2020-02-13
Added support for disk performance optimization.

garvct:
Add support for premium disk performance optimization (tag to try and keep attached disks and VM close to each other.)


libexec/azhpc-build.sh
```diff
@@ -300,6 +301,11 @@ for resource_name in $(jq -r ".resources | keys | @tsv" $config_file); do
                 fi
                 data_disks_options=
                 if [ "$resource_disk_count" -gt 0 ]; then
+                    disk_perfoptlevel_option=
+                    if [ "$resource_disk_perfoptlevel" = "true" ]; then
+                        status "Enable disk performance optimization"
+                        disk_perfoptlevel_option="--tag '\$perfOptimizationLevel=1'"
```
xpillons: Is there more values that 1 for this tag ? if so please consider passing it thru the variable itself

garvct: Thanks Xavier, there is only one value for tag that I am aware of. (That may change in the future?).

PR #194, opened 2020-04-01
Added Grafana

xpillons:
Added Grafana and telegraf for monitoring resources
 - installation scripts for grafana
 - configure grafana database
 - initialize InfluxDB datasoure
 - import default dashboard
 - installation script for telegraf (monitoring agent)


scripts/install-telegraf.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/sh
```
edwardsp, line 1: This should be "#!/bin/sh" (or #!/bin/bash) - it's not valid on my wsl2 or ubuntu linux.

scripts/install-grafana.sh
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/sh
```
edwardsp, line 1: invalid path - see other comment

scripts/install-telegraf.sh
xpillons: update done

PR #199, opened 2020-04-03
Js convergcfd

JonShelley:
Initial add for ConvergeCFD.


apps/convergecfd/run_ccfd_2.4_impi.pbs
```diff
@@ -0,0 +1,85 @@
...
+NODES=$(cat $PBS_NODEFILE | sort -u | wc -l)
+
+export RLM_LICENSE=$LICENSE_INFO
+export PATH=${APP_INSTALL_DIR}/convergecfd/$CCFD_VERSION/l_x86_64/bin:$PATH
+
+# Setup Intel MPI
+module load mpi/impi-2019
+source $MPI_BIN/mpivars.sh -ofi_internal
+export I_MPI_FABRICS="shm:ofi"
+export I_MPI_FALLBACK_DEVICE=0
```
xpillons, line 26: I_MPI_FALLBACK_DEVICE can be removed when using OFI as it is no longer supported

apps/convergecfd/run_ccfd_3.0_impi.pbs
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
```
xpillons, line 1: Would it be possible to make a single script for both versions ? there is a lot of duplicated lines which can end into errors

apps/convergecfd/run_ccfd_3.0_impi.pbs
```diff
@@ -0,0 +1,91 @@
...
+
+#PBS -N ConvergeCFD
```
xpillons, line 3: As all the PBS options are specified thru the QSUB can these lines be removed and the script renamed from .pbs to .sh ?

apps/convergecfd/run_ccfd_2.4_impi.pbs
JonShelley: I have removed I_MPI_FALLBACK_DEVICE from the script

apps/convergecfd/run_ccfd_3.0_impi.pbs
JonShelley: It's possible but I am not sure of the value for the additional work. Many things have changed between the versions and 2.4 is no longer be actively developed. I only added the 2.4 run script for those companies that have not switched over to 3.0 yet. If you would only rather to have a single run script for 3.0 then I can remove the impi extension on the filename and do some additional work when their HPCX version becomes available so that I don't need to have a run_ccfd_3.0_hpcx.pbs version

apps/convergecfd/run_ccfd_3.0_impi.pbs
JonShelley: I would rather leave it as a .pbs file extention and leave in the #PBS lines since they are comments anyways and make it easier to submit the job when using PBS. Also the script calls out $PBS_O_WORKDIR, $PBS_NODEFILE, $PBS_HOSTFILE, and $PBS_JOBID so it would not work for another workload manager. If it needed to be run in SLURM or LSF then we would need to change those values in order to get it to work.

PR #205, opened 2020-04-06
Changed vm os disk size default.

garvct:
The current vm os disk size default does not work deploying a Windows server (e.g example anf_cifs/ad-config.json). The windows server requires an os disk >= 128 GB (The current default of 32 GB causes a deployment failure).

In this proposed PR, --os-disk-size is only passed if os disk size is explicitly set in the config file, otherwise a suitable default is set automatically.


edwardsp:
The change is fine @garvct but we should all be using the python version now.  @xpillons and I were talking about removing the bash version yesterday as we now have features there not available in bash.  Please try the python version (remove your bin directory and run the install again) and report any issues and I'll fix them.

garvct:
Ok, will close this pull request and work on python version. Thanks,

PR #210, opened 2020-04-07
Corrected anf_cifs/ad-config.json

garvct:
The python version currently does not support deploying without a jumpbox/headnode (The ad-config.json technically does not require a jumpbox because it only executes a local_script).

-Added a jumpbox (work-around)
-Changed post-install script to local_script  (Should we support post-install in the python version for backward compatibility?)

### DIFF
--- a/examples/anf_cifs/ad-config.json
+++ b/examples/anf_cifs/ad-config.json
@@ -1,10 +1,9 @@
 {
   "location": "variables.location",
   "resource_group": "variables.resource_group",
-  "install_from": "jumpbox",
+  "install_from": "",
   "admin_user": "variables.ad_admin",
   "variables": {
-    "image": "OpenLogic:CentOS:7.7:latest",
     "location": "westeurope",
     "resource_group": "ad_demo",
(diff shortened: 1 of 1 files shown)


xpillons:
> The python version currently does not support deploying without a jumpbox/headnode (The ad-config.json technically does not require a jumpbox because it only executes a local_script).
> 
> -Added a jumpbox (work-around)
> -Changed post-install script to local_script (Should we support post-install in the python version for backward compatibility?)

we won't add post-install in the python as it's replaced by local_script. it's a breaking change, but it's more powerfull and need a simple config file update.

garvct:
Thanks, agreed, building without a jumpbox would be the preferred option (avoid wasting resources), will re-run test (removing jumpbox) with updated version when available.

edwardsp:
Please can you test this again without a jumpbox @garvct ?

garvct:
Yes, will run a test and let you know.

garvct:
I removed the jumpbox, ad-config deploys correctly.

PR #214, opened 2020-04-08
Changed os and data disk defaults to Premium SSD

garvct:
Changed the default OS and data disk to Premium SSD's so there are no performance surprises when an NFS server or OS disk is created with default settings.

### DIFF
--- a/pyazhpc/arm.py
+++ b/pyazhpc/arm.py
@@ -389,9 +389,9 @@ class ArmTemplate:
         ran = res.get("accelerated_networking", False)
         rlowpri = res.get("low_priority", False)
         rosdisksize = res.get("os_disk_size", None)
-        rosstoragesku = res.get("os_storage_sku", "StandardSSD_LRS")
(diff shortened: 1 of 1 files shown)


garvct:
I think Premium_SSD's will typically deliver better performance than Standard SSD's. For Production environments/benchmarking I think Premium SSD's would be preferred.
If we leave Standard SSD's as the default at minimum I would like to see it as a mandatory option in the config file, so its clear that Standard SSD's are used and not Premium SSD's.

garvct:
re-opened.

edwardsp:
Shall we discuss and decide as a group on this one next team meeting?

garvct:
My preference would be to create a separate PR updating config files explicitly including the disk type parameter. Would that be ok?

xpillons:
> My preference would be to create a separate PR updating config files explicitly including the disk type parameter. Would that be ok?

Sure

PR #215, opened 2020-04-09
Support custom image

garvct:
Added support for Custom images.

### DIFF
--- a/pyazhpc/arm.py
+++ b/pyazhpc/arm.py
@@ -5,8 +5,6 @@ import uuid
 import azlog
 import azutil
 
-import re
-
 log = azlog.getLogger(__name__)
 
 class ArmTemplate:
@@ -360,7 +358,7 @@ class ArmTemplate:
         return datadisks
 
     def __helper_arm_create_image_reference(self, refstr):
-        if re.search(":", refstr):
+        if ":" in refstr:
            return {
               "publisher": refstr.split(":")[0],
               "offer": refstr.split(":")[1],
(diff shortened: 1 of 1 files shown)


pyazhpc/arm.py
```diff
@@ -358,11 +360,16 @@ def __helper_arm_create_datadisks(self, sizes, sku, cache):
...
-            "sku": refstr.split(":")[2],
-            "version": refstr.split(":")[3]
+        if re.search(":", refstr):
+           return {
+              "publisher": refstr.split(":")[0],
+              "offer": refstr.split(":")[1],
+              "sku": refstr.split(":")[2],
+              "version": refstr.split(":")[3]
+        }
+        else:
```
xpillons, line 370: Are you sure that you can't use : in a custom image name ? If so this is going to break this, otherwise this is good.

pyazhpc/arm.py
```diff
@@ -5,6 +5,8 @@
 import azlog
 import azutil
 
+import re
```
edwardsp, line 8: no need for re here

pyazhpc/arm.py
```diff
@@ -358,11 +360,16 @@ def __helper_arm_create_datadisks(self, sizes, sku, cache):
         return datadisks
 
     def __helper_arm_create_image_reference(self, refstr):
-        return {
-            "publisher": refstr.split(":")[0],
-            "offer": refstr.split(":")[1],
-            "sku": refstr.split(":")[2],
-            "version": refstr.split(":")[3]
+        if re.search(":", refstr):
```
edwardsp, line 363: you can just do:
if  ":" in refstr:

PR #230, opened 2020-04-20
Spack update (v0.14.2)

garvct:
Updated spack to v0.14.2 (previous version was 0.13.1)
Updated to latest CentOS-HPC 7.7 image (e.g openmpi is now 4.0.3 etc)


apps/spack/build_spack.sh
```diff
@@ -23,7 +23,6 @@ git checkout tags/v${APP_VERSION}
 
 mkdir ${SPACKDIR}/spack/var/spack/repos/builtin/packages/hpcx
 cp ${APPS_SPACK_DIR}/package.py  ${SPACKDIR}/spack/var/spack/repos/builtin/packages/hpcx
```
xpillons, line 25: How is APPS_SPACK_DIR set ? I don't see it in the that script.

PR #231, opened 2020-04-20
WRF update (to 4.1.5, previous version 4.1.3)

garvct:
Update WRF to version 4.1.5
-Some minor changed due to spack update to 0.14.2)


garvct:
Thanks Xavier, there is also a dependency on the spack version (i.e The old WRF 4.1.3 will install with 0.13.1 but not 0.14.2 (without changes)). It someone wants to build WRF 4.1.3, I would recommend that  they build it with the new spack and latest openmpi library on our Marketplace image (not an older version))

PR #233, opened 2020-04-21
Increased saskey expiration to 2 hours

garvct:
I have had a number of deployment errors because generated saskey expired.

Would like to increase expiration to 2 hours.


xpillons:
Does your scripts runs for more than 1h ? that would be the only case then.

garvct:
Yes, my deployment can take over an hour. Also, when you have a failure that requires sometime to debug, I often cannot manually re-run some of the failed scripts because the generated saskey has expired and I need to re-generate them.

PR #234, opened 2020-04-21
make_filesystem modification

garvct:
Changed permission of mounted filesystem.

When using make_filesystem.sh to mount HB120rs_v2 nvme local ssh, I could not write because I did not have permission.


garvct:
Middle ground may be to have the permissions as an arg to the script (Setting conservative default permissions).

PR #249, opened 2020-04-28
Glusterfs

garvct:
Initial support for glusterfs in azurehpc.


scripts/glusterfs_vol.sh
```diff
@@ -0,0 +1,33 @@
...
+HOSTLIST=$1
+STRIPE=$2
+REPLICA=$3
+
+GLUSTERFS_VOL_NAME=gv0
+
+mkdir -p /mnt/brick1/$GLUSTERFS_VOL_NAME
+
+if [ "$PSSH_NODENUM" = "0" ]; then
+   for host in `cat ~hpcadmin/azhpc_install_config/hostlists/$HOSTLIST`
```
xpillons, line 12: config file name is hardcoded as well ad the username, it's fine right now, but this can easily be broken by using another config file name or user

scripts/glusterfs_vol.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+HOSTLIST=$1
+STRIPE=$2
+REPLICA=$3
+
+GLUSTERFS_VOL_NAME=gv0
+
+mkdir -p /mnt/brick1/$GLUSTERFS_VOL_NAME
```
xpillons, line 9: /mnt/brick1 is hardcoded, can you add it as a param and default to that value ?

scripts/glusterfs_pool.sh
```diff
@@ -0,0 +1,13 @@
+#/bin/bash
+
+HOSTLIST=$1
+
+systemctl start glusterd
+systemctl status glusterd
+
+if [ "$PSSH_NODENUM" = "0" ]; then
+   for host in `cat ~hpcadmin/azhpc_install_config/hostlists/$HOSTLIST`
```
xpillons, line 9: username and config file name hardcoded, maybe the hostlist full path should be provided instead ? @edwardsp what do you think ?

edwardsp: the scripts are run from inside the temp directory - you could just use "cat hostlists/$HOSTLIST".  But, better still, you could just set the arg in the config as "$(<hostlists/glusterfs)" and then you would get it as the first arg.  You would only then need to iterate as follows:
for host in $HOSTLIST; do ...

scripts/glusterfs_vol.sh
garvct: Thanks Xavier, I changed the example dir name to be glusterfs_ephemeral.

scripts/glusterfs_vol.sh
garvct: Added BRICK_MOUNT_PT variable with default /mnt/brick1

scripts/glusterfs_pool.sh
garvct: Thanks Paul and Xavier, I made the changes you suggested, looks cleaner now.

PR #255, opened 2020-04-30
Slurm in azhpc

edwardsp:
This is a first version but a working example.  Still more work to do including:

* Tidying up where to install
* Multiple queues
* Supporting more VM types


examples/slurm_autoscale/scripts/hpcadmin_id_rsa
```diff
@@ -0,0 +1 @@
+../hpcadmin_id_rsa
```
xpillons, line 1: probably this is not needed

examples/slurm_autoscale/scripts/hpcadmin_id_rsa.pub
```diff
@@ -0,0 +1 @@
+../hpcadmin_id_rsa.pub
```
xpillons, line 1: probably this is not needed

examples/slurm_autoscale/scripts/munge.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
xpillons, line 1: can all these scripts be moved to the azhpc/scripts directory ?

examples/slurm_autoscale/scripts/config.json
```diff
@@ -0,0 +1 @@
+../config.json
```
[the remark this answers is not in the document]
edwardsp: they are symbolic links in order to copy to the headnode

examples/slurm_autoscale/scripts/hpcadmin_id_rsa
edwardsp: see above

examples/slurm_autoscale/scripts/hpcadmin_id_rsa.pub
edwardsp: see above

PR #270, opened 2020-05-08
Azure Bastion example (No Public IP access)

garvct:
Azure bastion example, setting up a HPC cluster with no public IP access. 
Access Windows or Linux VM in the private VNET using Azure bastion.


lmiroslaw:
Any chances to merge this example into master branch? I have a customer project where we need Bastion to connect to Linux VM

xpillons:
@lmiroslaw can you describe the customer scenario ? is it about doing all deployment thru the Bastion ? adding a bastion access after a deployemnt have been done thru a jumpbox with a public IP ?

lmiroslaw:
We use Azure Bastion (with public IP but behind RBACs) to connect to a jumpbox VM over HTTPS. From the jumpbox we want to browse or CLI to the CycleCloud  and SSH into any node as needed.

garvct:
I will add a second config file to create a bastion and jumpbox (private IP) with the intention of deploying the rest of your resources from the jumpbox.

garvct:
I added a second config file (config_no_pub_ip.json), which will deploy the Azure bastion, VNET and a jumpbox(No pub IP). The idea would be to use the jumpbox to deploy the rest of the azurehpc deployment. (This would work for sites which have a strict policy of no pub IP's allowed).

PR #277, opened 2020-05-13
single node health checks

xpillons:
single node IB device checks
can be easily extended to new SKUs thru the JSON dictionary file
need to be extended to add stream


xpillons:
Sure. apps/health_checks is different and not easily applicable on a deployment, pbshooks are specific to PBS and too deep integrated to be able to be reused in another context. That script could be called from the pbshook for example or by any other mechanism.

garvct:
Looks good to me.

PR #289, opened 2020-06-05
Lustreinfiniband

chadnar2:
- lustre-ipoib - This is a created implementation of Lustre using ip over infiniband (IPoIB)
- lustre-rdma - This is a created implementation of Lustre using native Remote Direct Memory Access (RDMA)

lustre_rdma_nvmedrives:
Changes to files to enable Infiniband functionality:
lfsmaster.sh
lfsoss.sh
lfsclient.sh
lfsrepo.sh
lfspkgs.sh

Addition for the installation of new Mellanox OFED (MOFED) for the Lustre kernel : installMOFED.sh

Addition for correct drives placement of OSSes : installdrives.sh
*installdrives.sh takes about 15 minutes to run so please either remote this entity, or wait it out.

Additions for correct Lustre kernel :
lustreinstall1.sh
lustreinstall2.sh

Addition for pause after MDS/OSS reboot : waitforreboot.sh

### DIFF
--- a/examples/lustre_ipoib/config.json
+++ b/examples/lustre_ipoib/config.json
@@ -1,0 +1,223 @@
+{
+    "location": "southcentralus",
+    "resource_group": "variables.resource_group",
+    "install_from": "headnode",
+    "admin_user": "hpcadmin",
+    "vnet": {
+        "name": "hpcvnet",
+        "address_prefix": "10.2.0.0/20",
+        "subnets": {
+            "compute": "10.2.0.0/22",
+            "storage": "10.2.4.0/24"
+        }
+    },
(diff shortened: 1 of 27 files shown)


chadnar2:
I have reduced the size of the headnode to a 'Standard_D8s_v3' since there is no infiniband connectivity with the Lustre servers anyway. 
These are not HPC images, hence the need to install OFED. MOFED never worked for IB Lustre when we tried to use the HPC image. This may be something to look at.
The waagent restart comes from another Lustre script.
What do you suggest to actively check if the Lustre kernel has been installed and the node is up? (ssh -q hpcuser@hostname exit ??)

PR #293, opened 2020-06-13
Change fault domain count to be 1

garvct:
The default fault domain count is 5.

Large VMSS (300 HC instances in south central us) have failed to deploy when fault domain count is set to 5, but deploys correctly when set to 1.


xpillons:
@garvct does settings the FD to 5 in the config file can solve this ? if so there is no need to override the default value to 1.

garvct:
yes, setting the fault domain count to 1 in the the config file fixes the problem. Why not set the default  to 1 to prevent other Users from having deployment errors. What is the downside of setting the default domain count to be 1?

PR #302, opened 2020-06-24
Support for UltraSSD DataDisk

xpillons:


xpillons:
@hmeiland can you please unit test that ?

pyazhpc/arm.py
```diff
@@ -651,7 +654,10 @@ def _add_vm(self, cfg, r, vnet_in_deployment):
                 vmres["properties"]["proximityPlacementGroup"] = {
                     "id": f"[resourceId('Microsoft.Compute/proximityPlacementGroups','{rppgname}')]"
                 }
-            
+
+            if rstoragesku == "UltraSSD_LRS" :
+                vmres["properties"]["additionalCapabilities"] = { "ultraSSDEnabled": "true" }
```
edwardsp, line 659: is this supposed to be a string or bool value?

edwardsp: @hmeiland 
fyi here is the rest api - https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/createorupdate#additionalcapabilities - which looks to be of type bool

hmeiland: yep: agree

hmeiland:
> you need to set as **True** (with a capital T) rather than **true** as this is in Python

yes, updated with newer commit

xpillons:
@edwardsp @hmeiland this is all tested now with the latest fix about zones and public IP. pipeline passed so it can be approved

PR #308, opened 2020-06-25
Create run_benchmark20m-2x120cores.pbs

chadnar2:
This is for openFoam on HBv2


xpillons:
and finally i think it's not needed as there is already a generic run_benchmark.sh script so even better, you can delete both .pbs files

chadnar2:
This was for HBv2 with 120 cores per node. I can take a look later today at the differences as this file is old.

chadnar2:
So just delete this branch?

xpillons:
yes

PR #309, opened 2020-06-25
Deploy CycleCloud Slurm cluster with BeeGFS

vanzod:
**Changelog:**
- Added scheduler selection in `variables.json` and corresponding logic in `init.sh`
- Corrected `slurmcycle.json` template
- Added scripts to run tests in Slurm
- Updated documentation


apps/linpack/single_hpl.sh
```diff
@@ -82,4 +82,4 @@ $HPL_NB           swapping threshold
 8            memory alignment in double (> 0)
 EOF
 
-mpirun $mpi_options $HPL_EXE | tee hpl.out
+mpirun $mpi_options $HPL_EXE
```
xpillons, line 85: please keep the tee as it's used by the parsing script so that we can load these values in log analytics

apps/ior/ior.slurm
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
xpillons, line 1: This script is the same than the ior.sh, can the #SBATCH lines be added on the SLURM command line so we keep a single script and not one per job scheduler ?

vanzod: This is actually one thing I wanted to ask. For simplicity of testing I went ahead and created separate scripts for Slurm, but that generates one additional file that needs to be maintained. I can work on adding the necessary logic in the current bash scripts to correctly manage both PBS and Slurm.

vanzod:
Now the same test scripts can be used on either PBS or Slurm cluster.
All scripts have returned the correct results on both PBS and Slurm clusters.

PR #311, opened 2020-06-29
JSON was only assumed and could be overridden

edwardsp:


edwardsp:
> Is it the only place where you need to parse content in json format for the AZ CLI ?

it's the only place I could see - we tend to ask for tsv when we want specific values

PR #319, opened 2020-07-01
PBS automatic deletion of idle vmss instances

garvct:
The pbs_delete_idle_vmss.sh script generated a script, runs the script at regular intervals (default 10min, can be changed in the config.json) defined by the crontab.

- Runs on headnode (no ssh to compute instances, looks at the PBS resource "last_used_time" and job "status")
- When a PBS job has not run on the VMSS instance for the specified period. The instance will be removed from PBS and deleted.
- Will work with multiple VMSS's
- The instance idle time measurement only starts when a PBS job has completed on the instance. (i.e When the HPC cluster is first deployed the idle measurement does not start).
- A logfile (/tmp/azurehpc_delete_idle_vmss.log_$$) is generated to monitor/debug/record idle instances and what instances have been deleted.

### DIFF
--- a/examples/pbs_delete_idle_vmss/config.json
+++ b/examples/pbs_delete_idle_vmss/config.json
@@ -1,0 +1,156 @@
+{
+  "location": "variables.location",
+  "resource_group": "variables.resource_group",
+  "install_from": "headnode",
+  "admin_user": "hpcadmin",
+  "variables": {
+    "image": "OpenLogic:CentOS-HPC:7.7:latest",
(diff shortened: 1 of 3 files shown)


garvct:
The simple_hpc_pbs example is a basic/simple example to get started with AzureHPC. It should be kept simple? This is a new feature/capability that not everyone will want to use, the example config has 2 VMSS's, azcli and pbs_delete_idle_vmm script (and specific prerequisites). I feel it would be better and clearer for users  to have its own example.

xpillons:
ok, I miss that part. make sense then to have this dedicated.

PR #320, opened 2020-07-03
Added Cloud-init support in AzureHPC 

garvct:
Added support for cloud-init in AzureHPC, primary use case would be install software on no-public IP VM/VMSS
(e.g AzureHPC prerequisites (git, jq, AzureHPC git clone, azcopy and azcli) (see cloud-init.txt example file).

Cloud-init scripts need to be uploaded to blob storage in advance of AzureHPC deployment.

The Cloud-init scripts can be then referenced using "cloud_init_sasurl" key.

e.g "cloud_init_sasurl": "sasurl.{{variables.storage_acc}}.data/cloud-init.txt"

The bastion config_no_pub_ip.json was extended to include cloud-init (Set-up AzureHPC prerequisites on no-pub IP jumpbox).

### DIFF
--- a/examples/bastion/config_no_pub_ip.json
+++ b/examples/bastion/config_no_pub_ip.json
@@ -15,7 +15,7 @@
         "storage_acc": "<NOT-SET>",
         "vnet_resource_group": "variables.resource_group",
         "jumpbox_pw": "secret.{{variables.key_vault}}.jumpbox_pw",
-        "cloud_init_sasurl": "sasurl.{{variables.storage_acc}}.data/cloud-init.txt"
+        "customdata": "sasurl.{{variables.storage_acc}}.data/cloud-init.txt"
     },
     "vnet": {
         "resource_group": "variables.vnet_resource_group",
@@ -33,7 +33,7 @@
             "public_ip": false,
             "image": "variables.jumpbox_image",
             "accelerated_networking": true,
-            "cloud_init_url": "variables.cloud_init_sasurl",
+            "custom_data": "variables.customdata",
             "subnet": "compute",
             "tags": []
         }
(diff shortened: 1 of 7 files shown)


examples/bastion/config_no_pub_ip.json
```diff
@@ -30,6 +33,7 @@
             "public_ip": false,
             "image": "variables.jumpbox_image",
             "accelerated_networking": true,
+            "cloud_init_sasurl": "variables.cloud_init_sasurl",
```
edwardsp, line 36: This doesn't need to be a sasurl - any url is fine, so it might be better to just name cloud_init or cloud_init_url?

edwardsp:
FYI cloud init is supported in azure images from centos 7.7 onwards.

examples/bastion/config_no_pub_ip.json
garvct: I changed the cloud_init keyword to cloud_init_url and updated the changelog.

examples/bastion/config_no_pub_ip.json
```diff
@@ -4,15 +4,18 @@
     "install_from": "",
     "admin_user": "hpcadmin",
     "variables": {
-        "jumpbox_image": "OpenLogic:CentOS:7.7:latest",
+        "jumpbox_image": "OpenLogic:CentOS-CI:7-CI:latest",
```
edwardsp, line 7: This image should not be used - see details here: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/using-cloud-init#centos

garvct:
Thanks Paul, I made some the the changed you suggested, I also added the "#include" syntax automatically if a URL or SASURL is used to reference a Cloud-Init script.

PR #321, opened 2020-07-03
Only call deploy on build if resources exist

edwardsp:


xpillons:
Please keep creating the RG even if no resources exists

edwardsp:
> please keep creating the RG even if no resources are in the config file as local scripts can create additional resources like keyvault

the latest commit addresses this

PR #325, opened 2020-07-09
cc_anf example

hmeiland:
blocks and example for anf on CycleCloud; with dynamic anf naming based on uuid to protect for naming conflicts on anf.


vanzod:
CC+ANF now fully supports both Slurm and PBS.
Both clusters passed all tests.

PR #327, opened 2020-07-10
ND40rs v2 gpu numa mapping

garvct:
This calculates the correct ND40rs_v2 GPU mapping (topology and mapping to the correct numa domain).

I have complemented some existing scripts by adding a cuda bandwidthTest benchmark to determine the correct GPU mapping to host numa domain (or socket). The mapping file is written to /tmp/gpu_map_file  (e.g 3,4,5,6,7,2,1,0).

I also added a second gpu_mapping script called (bwtest_gpu_map.sh), this scripts used cuda bandwidthtest exclusively to determine the GPU mapping. It does a BW test from each GPU tp core 0-1 (Numa domain 0), then sorts the results and writes the mapping to /tmp/gpu_mapfile and the GPU BW results to /tmp/gpu_bwtest_file (good sanity check).

### DIFF
--- a/examples/gpu_numa_mapping/config.json
+++ b/examples/gpu_numa_mapping/config.json
@@ -1,0 +1,129 @@
+{
+  "location": "variables.location",
+  "resource_group": "variables.resource_group",
+  "install_from": "headnode",
+  "admin_user": "hpcadmin",
+  "variables": {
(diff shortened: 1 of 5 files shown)


garvct:
I have remove PBS and azcopy from the config.json.

PR #331, opened 2020-07-17
This is a working version of glusterfs with Cycle Cloud

chadnar2:
The copy of glusterfs_pool.sh inside of azurehpc was not used as it would not function with what had to be done. 
Others can see if they can get this to work.
A copy of glusterfs_pool.sh along with glusterfs_pool_master.sh has been included here. Everything was tested out.

### DIFF
--- a/blocks/glusterfs-cluster.json
+++ b/blocks/glusterfs-cluster.json
@@ -1,0 +1,148 @@
+{
+  "location": "variables.location",
+  "resource_group": "variables.resource_group",
+  "install_from": "jumpbox",
+  "admin_user": "hpcadmin",
+  "vnet": {
+    "name": "hpcvnet",
+    "address_prefix": "10.2.0.0/20",
+    "subnets": {
+    }
+  },
+  "variables": {
+    "resource_group": "<NOT-SET>",
+    "location": "<NOT-SET>",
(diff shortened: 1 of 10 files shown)


garvct:
Did these changes impact examples/glusterfs_ephemeral and examples/glusterfs_data_disks? Do they still work?

garvct:
We need one common set of scripts for all glusterfs examples?

chadnar2:
No changes to examples/glusterfs_ephemeral and examples/glusterfs_data_disks.

garvct:
You have a scripts dir in cc_glusterfs? all glusterfs examples need to use the glusterfs scripts in azurehpc/scripts. If these scripts are not working then they need to be modified not create new scripts in cc_glusterfs/scripts.

garvct:
Please restore the branch until this is corrected.

chadnar2:
We'll have to dig into whether examples/glusterfs_ephemeral and examples/glusterfs_data_disks use the existing glusterfs_pool.sh. If so, something needs to change. I had to keep the head glusterfs node up for cyclecloud and this was the approach taken for other cyclecloud builds. The glusterfs_pool.sh did not seem to function as designed, so it was broken into 2 scripts (one for master, one for clients). We should talk about the approach then.

PR #341, opened 2020-08-04
glusterFS example with CycleCloud

xpillons:
- GlusterFS with Availability Set building block
- CycleCloud PBS and SLURM cluster support


garvct:
I assume being in the same avset guarantees that these resources are in the same datacenter (and its not necessary to also specify they are in the same proximity placement group).?

xpillons:
correct

PR #342, opened 2020-08-04
added pipeline for CC_ANF

xpillons:
- added pipeline
- updated doc with badge
- updated changelog
- updated instructions


examples/cc_anf/pipeline.yml
```diff
@@ -0,0 +1,42 @@
...
+    - master
+    - '*'
+  paths:
+    include:
+      - examples/cc_anf/*
+      - blocks/cycle-install-server-managed-identity.json
+      - blocks/cycle-prereqs-managed-identity.json
+      - blocks/cycle-cli-local.json
+      - blocks/cycle-cli-jumpbox.json
+      - blocks/jumpbox-anf.json.json
```
vanzod, line 15: I believe there is an extra `.json` here

xpillons: good catch

PR #349, opened 2020-08-27
Eda

mengruts:
Compiled 4 EDA tools on \apps:
cadence_clearity3d
cadence_tempus
cadence_spectrex
cadence_xcelium


hmeiland (approved):

xpillons:
@mengruts please see my comments, especially consider your script to be able to download the binaries from another blob and not from yours.

PR #350, opened 2020-08-31
Update to Spack 0.15.4 and added support azure blob storage as an spack buildcache

garvct:
-Update spack to 0.15.4
-Corrected readme documentation.
-Added support for Azure blob storage as an spack buildcache.
format is  azure://<STORAGE_ACCOUNT_NAME>.blob.core.windows.net/
(Will submit a PR to spack after this review).


apps/spack/package.py
```diff
@@ -31,7 +31,7 @@ class Hpcx(AutotoolsPackage):
     provides('mpi')
                                                                                                    
     def setup_dependent_environment(self, spack_env, run_env, dependent_spec):
-        self.prefix.bin = "/opt/hpcx-v2.5.0-gcc-MLNX_OFED_LINUX-4.7-1.0.0.1-redhat7.7-x86_64/ompi/bin"
+        self.prefix.bin = "/opt/hpcx-v2.6.0-gcc-MLNX_OFED_LINUX-5.0-1.0.0.0-redhat7.7-x86_64/ompi/bin"
```
xpillons, line 34: Can this be dynamically discovered based on the current HPC image you are running on ? We probably don't want to change this file each time the image is upgraded.

apps/spack/azure_blob.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def azure_upload_to_blob(self, local_file_path):
+        from azure.storage.blob import ContentSettings
+        if Path(self.blob_path).suffix == '.json':
+           contentsettings = ContentSettings(content_type="application/json")
+        else:
+           contentsettings = ContentSettings()
+        blob_client = self.blob_service_client.get_blob_client(container=self.container_name, blob=self.blob_path)
+        with open(local_file_path, "rb") as data:
+            blob_client.upload_blob(data, overwrite=True, content_settings=contentsettings)
```
xpillons, line 78: Is there any retry logic here ? if not then there should be one. 
how does errors are handled and recovered ?

apps/spack/package.py
garvct: ok thanks, I have updated package.py to get the path from the module file.

apps/spack/azure_blob.py
garvct: I am using an Azure Blob Storage python SDK, I would expect that it would handle some of the lower level re-try login you would need in RESP API? I have added some logic to catcc and report any failures.

xpillons: Retry policy is on by default see there https://docs.microsoft.com/en-us/azure/developer/python/sdk/storage/storage-blob-readme?view=storage-py-v12

xpillons (approved):

PR #352, opened 2020-09-02
Spack (Added azure_delete_blob and corrected azure_list_lob)

garvct:
-Added function azure_delete_blob()
-Corrected azure_list_blob
 Get correct index based on location of "build_cache".

### DIFF
--- a/apps/spack/azure_blob.py
+++ b/apps/spack/azure_blob.py
@@ -68,11 +68,15 @@ class AzureBlob:
 
 
     def azure_delete_blob(self):
-        blob_client = (self.blob_service_client.
-                       get_blob_client(container=self.container_name,
-                                       blob=self.blob_path))
-        blob_client.delete_blob()
-
+        try:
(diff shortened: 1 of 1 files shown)


apps/spack/azure_blob.py
```diff
@@ -67,6 +67,13 @@ def azure_blob_exists(self):
         return True
 
 
+    def azure_delete_blob(self):
+        blob_client = (self.blob_service_client.
+                       get_blob_client(container=self.container_name,
+                                       blob=self.blob_path))
+        blob_client.delete_blob()
```
xpillons, line 74: please add exception handling too

xpillons (changes requested):

apps/spack/azure_blob.py
garvct: Thanks, I added it.

xpillons (approved):

PR #353, opened 2020-09-03
Update fio to 3.22

garvct:
- Update fio to 3.22
-Corrected build script
-Added non-direct I/O tests
-Added flushing Linux buffer before I/O tests.

### DIFF
--- a/apps/fio/fio.pbs
+++ b/apps/fio/fio.pbs
@@ -1,4 +1,5 @@
 #!/bin/bash
+FILESYSTEM=${1:-$FILESYSTEM}
 set -o pipefail
 source /etc/profile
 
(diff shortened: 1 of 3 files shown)


apps/fio/fio.pbs
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-FILESYSTEM=$1
```
xpillons, line 2: Can you please keep the FILESYSTEM as a parameter as this is a breaking changed. you can initialize it by default to the FILESYSTEM env variable. more generally speaking if you expect an env variable to be set make sur it's defined and exit on error if not.

xpillons (changes requested):

apps/fio/fio.pbs
garvct: ok, thanks, I added the FILESYSTEM back as a parameter, setting the default to be $FILESYSTEM, so it can be set via environmental variable, PBS -v or command line arg.

xpillons (approved):

PR #355, opened 2020-09-07
Experimental use of Ansible for VM configuration

xpillons:
- using Ansible to configure VMs
- added roles for 
  - NFS Server
  - NFS Mount
  - PBS Server
  - PBS Client
  - Common
- Local script to automatically build Playbook
- direct mapping between azhpc tags and Ansible roles


btravouillon:
You should replace hyphen with underscore in role names, it will definitely help the day you want to migrate some roles to [collection roles which don't accept hyphen in the role name](https://docs.ansible.com/ansible/devel/dev_guide/migrating_roles.html#migrating-a-role-to-a-collection).

xpillons:
> You should replace hyphen with underscore in role names, it will definitely help the day you want to migrate some roles to [collection roles which don't accept hyphen in the role name](https://docs.ansible.com/ansible/devel/dev_guide/migrating_roles.html#migrating-a-role-to-a-collection).

Thank you @btravouillon will definitively do it

edwardsp (approved):
Approved - we need to extend this further now and decide how to add more native support in azurehpc (e.g. maybe exporting the variable in azurehpc straight into ansible rather than doing by install steps).

PR #359, opened 2020-09-10
Added multiple SKU support and node oversubscription to Slurm Autoscale

vanzod:
The following improvements have been implemented:

1. User can specify multiple partitions in `config.json`, each targeting a different VM SKU.
2. Compute nodes are created upon job submission on availability sets dedicated for each SKU. This is necessary since multiple SKUs might not be available within the same availability set.
3. Jobs can be allocated on the same node if resources are available (aka node oversubscription).
4. CPU and memory on each node are uniquely allocated to a job via cgroups (no CPU/memory oversubscription).
5. On VMs with hyperthreading, only physical cores are allocated for jobs. In this way two separate jobs cannot share the same physical core. Within a job, both logical cores can be used.
6. 5% of the VM memory (up to a maximum of 5 GB) are reserved to system processes via `MemSpecLimit`.

Limitations:

1. SKUs must be specified at cluster creation since Slurm needs to know the exact type and number of nodes (even if with `state=CLOUD`) whenever the controller starts. To allow the selection of any SKU at job submission without specifying it in the `config.json` would require a controller restart each time a new VM SKU is added and the `slurm` user to have elevated privileges. Both requirements are not advisable.
2. Total memory available for a VM SKU and number of NUMA nodes must be known at cluster creation. Azure CLI currently does not report accurate values for the total memory and no information on the number of NUMA nodes. Hence the current approach depends on a SKU lookup table to determine the required values.


pyazhpc/azhpc.py
```diff
@@ -530,23 +534,17 @@ def do_slurm_resume(args):
...
-                resource_list.append(f"{resource_name}{part}")
-    else:
-        resource_list.append(resource_name)
-        resource_name = resource_name[:-4]
-    
+    resource_name, resource_list = _nodelist_expand(args.nodes)
+
+    # Generate node/sku lookup table from nodes.conf
+    sku_lookup = {}
+    for line in open('/apps/slurm/nodes.conf', 'r'):
```
edwardsp, line 541: we cannot force this location to be used

vanzod: @edwardsp I have kept the original file organization where all the Slurm configuration files are stored in `/apps/slurm` and then linked in `/etc/slurm`. A more robust approach would be to install the files directly in `/etc/slurm`.

edwardsp: I think `/apps/slurm` is fine for just one example (as we could create another config with a different location) but this is putting it into the azurehpc code.

vanzod:
@edwardsp I have removed the hardcoded path by using `scontrol` instead of parsing `nodes.conf`.
Since `azhpc` is executed on the VM that hosts the `slurmctld` daemon independently on where the job is actually submitted, `scontrol` will always be available.
The only slight downside of this approach is that for each resumed node a separate RPC is generated via `scontrol`. This might not be optimal in case the controller is under heavy load. This is a very remote corner case that will not affect the vast majority of deployments.

vanzod:
Added hyphen in node names as separator between resource name and node index.
If the resource name (aka partition) ends with one or more digit, prevents Slurm from considering them as part of the node index. This removes ambiguity when interpreting Slurm compact nodes nomenclature.

pyazhpc/azhpc.py
```diff
@@ -465,43 +466,51 @@ def wrapped_print(indent, text, max_width=80):
...
+                        nodes.append(f"{resource}{part}")
+                resource_names.append(resource)
             else:
-                nodes.append(resource_name)
-    
-    return nodes
+                nodes.append(resource)
+                resource_names.append(resource[:-4])
+
+    # Remove trailing hyphen if used in node names to separate
```
edwardsp, line 498: is there a case when this happens?  Azurehpc is currently coded to append a zero padded 4 digit number when creating multiple instances of a VM.

pyazhpc/azhpc.py
```diff
@@ -542,73 +551,62 @@ def do_slurm_resume(args):
...
-    deployname = azutil.deploy(
-        config["resource_group"],
-        output_template
-    )
-    log.debug(f"deployment name: {deployname}")
+        config["resources"] = {}
 
-    _wait_for_deployment(config["resource_group"], deployname)
+        # Iterate over all nodes which name starts with the resource name
+        for rname in filter(lambda x: x.startswith(resource), resource_list):
```
edwardsp, line 574: this would fail if you had two resource names where one is a substring of the other, e.g. viz and viz3d (which is currently the case for AzureHPC OnDemand)

edwardsp (changes requested):

pyazhpc/azhpc.py
```diff
@@ -542,73 +551,62 @@ def do_slurm_resume(args):
...
-    
-    config["resources"] = {}
-    for rname in resource_list:
-        config["resources"][rname] = template_resource
+    resource_names, resource_list = _nodelist_expand(args.nodes)
 
-    tpl = arm.ArmTemplate()
-    tpl.read_resources(config, False)
+    # Loop over all resources
+    for resource in resource_names:
```
edwardsp, line 557: We are going to end up doing multiple deployments here (one per resource type).  It would be better to create a single config with *all* the instances.

pyazhpc/azhpc.py
vanzod: I choose to add a hyphen in the node names to separate the resource name and the node index. The reason being that when the resource name ends with one or more digits (e.g. `d64dv4`), then when resuming multiple nodes, Slurm returns a node list in compact form and the digits in the resource name are added into the indexes (e.g. `d64dv[4001-4005]`), which causes `_nodelist_expand` to fail since the index has more than four digits.

pyazhpc/azhpc.py
vanzod: Good point. Let me make it more robust.

pyazhpc/azhpc.py
vanzod: Decided to keep hyphen separation in offline discussion. Resolving conversation.

pyazhpc/azhpc.py
vanzod: @edwardsp This is not an issue in the Slurm setup deployed here since each partition contains nodes belonging to the same resource and jobs cannot be submitted across multiple partitions.
That said, it remains a good point. I have modified the logic to correctly handle multiple resources in the same resume in case users manually add a partition that includes nodes belonging to multiple resources.

edwardsp (approved):
approved

PR #360, opened 2020-09-11
Slurm suspend fix

edwardsp:
VMs were not being deleted if multiple resource types were suspending in one call (e.g. interactive0001,viz3d0001)


xpillons (commented):
good to me, although should be tested in the slurm pipeline, I will add it

xpillons (approved):
should add tests in the slurm example later

PR #367, opened 2020-09-15
Added Weekly builds for examples

xpillons:
- added cron on all pipelines to build on weekends
- added a status page with all the pipeline badges of the examples


edwardsp (approved):
looks good

PR #370, opened 2020-09-16
support and test Gen2 images

xpillons:
- when capturing an image automatically detect the hyper-v generation so it doesn't need to be specified in the config file
- added multiple tests for building Gen1 and 2 GPU VMs
- added multiple tests for building and capturing Gen1 & Gen2 images


edwardsp (approved):
looks good

PR #377, opened 2020-09-17
Added python3 to cndefault

garvct:
-Added python3 to cndefault.sh

### DIFF
--- a/scripts/cndefault.sh
+++ b/scripts/cndefault.sh
@@ -9,7 +9,7 @@ do
     fi
 done
 
-yum -y install git jq htop
+yum -y install git jq htop python3
 
 # change access to resource so that temp jobs can be written there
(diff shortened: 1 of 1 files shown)


garvct:
python2 is now obsolete. Python (ie. python3) is expected to be available?
cndefault currently installs htop (is that required for all deployments?)

cndefault contains specific cyclecloud code (is that required for all deployments?)

xpillons:
if python2 is EOL I guess python3 will be added in images ?
I'm just cautious about the deployment time, htop is quick to install and is useful to troubleshooting, and CC code is only executed if needed. How much delay python3 installation will add here ?
if python3 is required for applications then it should be added as a deployment step and a dependency if needed.

garvct:
ok, thanks will add it elsewhere. Please close this PR.

PR #379, opened 2020-09-17
spack python3 requirement

garvct:
-Updated readme about spack python3 requirement.

### DIFF
--- a/apps/spack/readme.md
+++ b/apps/spack/readme.md
@@ -1,5 +1,8 @@
 # Spack
 
+## Prerequisites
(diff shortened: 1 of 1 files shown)


edwardsp (approved):

xpillons (approved):
would be good to add python3 in the spack install scripts too

garvct:
The spack install script does contain python3 install. Problem is only headnode needs to install spack, but when the compute nodes access spack they will fail because python3 is not installed. One easy solution would be to install python3 on all compute nodes with a tag.

PR #388, opened 2020-10-01
IO500 on BEEOND pipeline - experimental

xpillons:
- automated the IO500 benchmark on BEEOND
- added telemetry support in the pipeline
- improved PBS client installation
- improved startbeeond to failed in case of fallback to TCP if RDMA communication can't be established
- add a parameter to specify the number of metadata to be used
- fix ringpingpong parser
- upload pipeline deployment logs to blobs in case of failure and if a storage account is specified in the pipeline variables
-


edwardsp (approved):
looks good

PR #389, opened 2020-10-02
Grafana NSG and telegraf configuration file externalized

xpillons:
- add NSG for grafana in the config file options
- telegraf configuration file outside of the telegraf install script
- added global tags for vmsize and vmss name


edwardsp (approved):
looks good

PR #391, opened 2020-10-08
Spack + WRF + Darshan

xpillons:
**SPACK**

- refactored installation files
- create a spack module to initialize variables to run the environment script
- don't patch if already patched

**DARSHAN**

- wrapper to easily profile mpi apps
- use spack module

**WRF**
- use spack module
- failed in case of error when building
- pipeline to create a pbs cluster, build spack, build wrf, download new_conus12km, run a job with new_conus12km

**MISC**

- cleanup .gitignore
- added get_blob.sh script
- enhance ci scripts to submit pbsjob


apps/spack/build_spack.sh
```diff
@@ -24,33 +44,35 @@ git clone https://github.com/spack/spack.git
 cd spack
 git checkout tags/v${APP_VERSION}
 
-mkdir ${SPACKDIR}/spack/var/spack/repos/builtin/packages/hpcx
-cp ${APPS_SPACK_DIR}/package.py  ${SPACKDIR}/spack/var/spack/repos/builtin/packages/hpcx
```
garvct, line 28: Will this work? How does spack know about hpcx without defining it?

garvct (approved):

apps/spack/build_spack.sh
xpillons: this is just replacing the flat file copy by a file structure

PR #393, opened 2020-10-10
ZCentral remote Boost (Formerly HP RGS) on Linux VM

garvct:
E2E example of setting of Visualization on a Linux VM (NV12s_v3) using HPE ZCentral Remote Boost.
-Use extensions to enable M60 GPU's on NV12sv3
-Use XFce desktop
-Install ZCentral Remote Boost sender on linux VM (copied from blob storage)
-License file is also copied from blog storage
-readme contains images of process.
image


examples/zcentral_remote_boost_linux/config.json
```diff
@@ -0,0 +1,102 @@
...
+            "script": "linux_zcentral_remote_boost.sh",
+            "tag": "linux_zcentral_remote_boost",
+            "sudo": true,
+            "args": [
+               "variables.zcentral_remote_boost_sender_sas_url",
+               "variables.zcentral_remote_boost_sender_license_sas_url"
+            ]
+        },
+        {
+            "type": "local_script",
```
xpillons, line 71: This is not needed. You can add "nsg_allow": ["rdp"] in the config setting of the headnode.

examples/zcentral_remote_boost_linux/config.json
```diff
@@ -0,0 +1,102 @@
...
+    "install_from": "headnode",
+    "admin_user": "hpcadmin",
+    "variables": {
+        "hpc_image": "OpenLogic:CentOS:7_8:latest",
+        "location": "<NOT-SET>",
+        "resource_group": "<NOT-SET>",
+        "vm_type": "Standard_NV12s_v3",
+        "low_priority": false,
+        "vnet_resource_group": "variables.resource_group",
+        "zcentral_remote_boost_sender_sas_url": "<NOT-SET>",
```
xpillons, line 13: you can probably simplify this and use the sasurl.<STORAGE-ACCOUNT>.<STORAGE-PATH> syntax to let azhpc generate the sas key for you and have variables for storage-account and path. The syntax would be then :
sasurl.{{variables.storage_account}}.{{variables.storage_path}}

examples/zcentral_remote_boost_linux/config.json
```diff
@@ -0,0 +1,102 @@
...
+        },
+        {
+            "type": "local_script",
+            "script": "linux_nvidia_gpu_extension.sh",
+            "args": [
+                "variables.resource_group",
+                "headnode"
+            ]
+        },
+        {
```
xpillons, line 87: Why is this a local script ? can't it be executed directly on the headnode ?

examples/zcentral_remote_boost_linux/config.json
garvct: Removed this local script and added a ngs rule for zcentral (port 42966)

examples/zcentral_remote_boost_linux/config.json
garvct: I already assume that the sasurl macro will be used for "zcentral_remote_boost_sender_sas_url" and "zcentral_remote_boost_sender_license_sas_url"

examples/zcentral_remote_boost_linux/config.json
garvct: The linux_nvidia_gpu_extension.sh uses azcli to install the GPU extension, if I run it on the headnode I would need azcli installed (on headnode) and a service principle.

examples/zcentral_remote_boost_linux/config.json
xpillons: Great, then having the pattern to use in the value instead of NOT-SET will guide the end user on how to use it

examples/zcentral_remote_boost_linux/config.json
xpillons: this comment is not on the extension but on the restart_gdm.sh call

examples/zcentral_remote_boost_linux/config.json
garvct: Done.

examples/zcentral_remote_boost_linux/config.json
garvct: I need to restart gdm after the GPU extension, so I have to use a local script.

xpillons (approved):

PR #396, opened 2020-10-19
Added ISR11 hotfix

mengruts:


apps/cadence_spectre/build_spectre.sh
```diff
@@ -16,7 +16,7 @@ install_required_packages()
         sudo yum -y install ksh
```
xpillons, line 16: Do you need to call sudo on each installation line ? can't you just call the script in sudo instead ?

xpillons (approved):

PR #400, opened 2020-10-20
Removed sudo

mengruts:


mengruts:
@xpillons I removed sudo, and will fix other cadence scripts later on. Thanks.

xpillons (approved):

PR #403, opened 2020-10-21
Beeond 7.2

xpillons:
- upgrade to beeond 7.2
- added test matrix on HC, HB, HBv2
- removed PBS from the tests
- fix image version to 7.7.2020062600


garvct (approved):
Looks good to me.

PR #408, opened 2020-10-30
Updated run_ring_osu_bw.sh (health_checks)

garvct:
-update compiler version.
-update mvapich2 version.


xpillons (approved):
this is good , however there is a full set of script to run OSU BW in apps/osu supporting PBS and SLURM. It would be better to add mvapich2 support in them if any of interest

PR #413, opened 2020-11-09
beeond build fix, ibverbs API changed

GNSSYSTEMS-anla:


xpillons:
@garvct what do you think about this fix ?

garvct:
I attempted to build BeeOND with this modification on HBv2 (OpenLogic:CentOS-HPC :7.7:latest) (version 2020100700).

I got the following error

Building beegfs client module
/opt/beegfs/src/client/client_module_7/build/../source/common/net/sock/ibv/IBVSocket.c: In function ‘__IBVSocket_cleanupCommContext’:
/opt/beegfs/src/client/client_module_7/build/../source/common/net/sock/ibv/IBVSocket.c:714:24: error: void value not ignored as it ought to be
       int destroyRes = ib_destroy_cq(commContext->sendCQ);
                        ^
/opt/beegfs/src/client/client_module_7/build/../source/common/net/sock/ibv/IBVSocket.c:728:24: error: void value not ignored as it ought to be
       int destroyRes = ib_destroy_cq(commContext->recvCQ);
                        ^
make[3]: *** [/opt/beegfs/src/client/client_module_7/build/../source/common/net/sock/ibv/IBVSocket.o] Error 1
make[3]: *** Waiting for unfinished jobs....

What Azure HPC SKU's and OS version did you test this on?

Also, I think it would be a good idea (if you get this working) to put a check for the OFED version. i.e if ofed >= 5.1 then modify code.   You can easily get the ofed version using ofed_info -d

ofed_info -d
MLNX_OFED_LINUX-**5.1**-0.6.6.0:

Thanks.

GNSSYSTEMS-anla:
i used 5.1-2.4.6.0, i will add an if for that, also for the beegfs version. this was the only combination that worked for me at all.

garvct:
Thanks, can you also tell me what OS Version you were testing on (did you use a marketplace image or build your own custom image?)

GNSSYSTEMS-anla:
i used also OpenLogic:CentOS-HPC :7.7:latest and applied the changes here, because 7.8 was not available yet (maybe it is now) https://github.com/Azure/azhpc-images/tree/master/centos/centos-7.x/centos-7.8-hpc

PR #417, opened 2020-11-11
Install steps are now idempotent

edwardsp:
A marker is created **on each VM** when an install step is completed successfully.  It will only run if the marker is not present.

This has been tested with simple_hpc_pbs and works for scaling up the compute pool (manual steps are required to scale down).

Note: all logs are **appended to**.  This means that the output for previous successful runs will also be shown.  When the marker is detected it is reflected in the output.


xpillons (approved):
good for me

PR #420, opened 2020-11-17
Update run_ring_osu_bw_hpcx_pbs.sh

JonShelley:
Added the memory request. Useful if cgroups are enabled in PBS


garvct (approved):

JonShelley:
@garvct I just added one more file. Can you take a look and let me know if you are fine with that change as well?

garvct (approved):
I would suggest to remove the comments and also remove dropping the cache (it does not add any value ).

PR #429, opened 2020-12-14
Js ubuntu support

JonShelley:
Made changes to support Ubuntu. This is needed to support NDv2 and NDv4 fleet health checking

### DIFF
--- a/scripts/create_raid0.sh
+++ b/scripts/create_raid0.sh
@@ -18,6 +18,57 @@ while (( "$#" )); do
     shift
 done
 
+# Check to see which OS this is running on. 
+os_release=$(cat /etc/os-release | grep "^ID\=" | cut -d'=' -f 2 | sed -e 's/^"//' -e 's/"$//')
+os_maj_ver=$(cat /etc/os-release | grep "^VERSION_ID\=" | cut -d'=' -f 2 | sed -e 's/^"//' -e 's/"$//')
+echo "OS Release: $os_release"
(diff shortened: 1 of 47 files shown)


scripts/create_raid0.sh
```diff
@@ -12,6 +18,57 @@ while (( "$#" )); do
     shift
 done
 
+# Check to see which OS this is running on. 
+os_release=$(cat /etc/os-release | grep "^ID\=" | cut -d'=' -f 2 | sed -e 's/^"//' -e 's/"$//')
```
xpillons, line 22: It would be better to include a library script with a function to return the OS versions or with helpers like is_ubuntu or is_centos instead of duplicating all these lines in plenty of scripts. This will make it easier for maintenance. call the include script at the beginning and add it as a dependency in the config files where the script is used.

xpillons:
@JonShelley can you please split this PR into 2, one for OpenPBS and one with the ubuntu support. I would suggest you create a new branch for OpenPBS and update only files for it. Once done it would be easier to merge in the master.
once we have merge this OpenPBS in the master, you can merge the master in this ubuntu branch, and solve the conflicts.

xpillons (changes requested):

xpillons:
@JonShelley maybe the easiest would be to leave the jumpbox running on CentOS but allow compute nodes and job scheduler to run on Ubuntu. Doing this you need smaller updated in the python code. 
@edwardsp please comment on this too.

PR #430, opened 2020-12-15
imb-mpi1 (added bandwidth measurement support)

garvct:
- Add a third command argument (TYPE to select between default (**latency**) measurement and **bandwidth**)


xpillons (approved):
looks good for me

PR #431, opened 2020-12-15
MLC (max_bandwidth) update

garvct:
- Moved setting up hugepages on HB from install script to run script (remove hugepages on HB after running MLC)

- Renamed MLC run scripts.


xpillons (approved):
looks good for me

Issue #444, opened 2021-01-04
Tag resource group with the pipeline name


xpillons:

xpillons:
Closed by PR #445

PR #445, opened 2021-01-04
Tag resource group with pipeline context

xpillons:
add tags to the resource group created with :

- pipeline name
- job id
- build number

Closing #444


edwardsp (approved):
approved

PR #459, opened 2021-01-22
Updated lustre to 2.12.6

garvct:
-Updated lustre to 2.12.6
-Can use latest CentOS-HPC image and not have lustre client incompatibility issues with MOFED 5.1


xpillons:
Please change also the blocks/lustre-cluster.json and the cc_lustre example

xpillons (approved):
good for me

PR #464, opened 2021-02-04
Added support for ephemeral OS disks

edwardsp:


xpillons (changes requested):
Can you please add this new option in the template config.json file at the root of the repo which contains all the options ?

edwardsp:
> Can you please add this new option in the template config.json file at the root of the repo which contains all the options ?

done

xpillons (approved):

Issue #466, opened 2021-02-09
Support OpenPBS 20


xpillons:
Add support for OpenPBS 20 as part of supporting CentOS 8.1

xpillons:
closed by #467

Issue #425, opened 2020-11-30
Cannot create a cluster from CentOS HPC 8.1 Internal VHD


souvik-de:
**Describe the bug**
"azhpc-builld" fails at pbsserver.sh (step 09) when trying to create a cluster from an Internal CentOS 8.1 HPC VHD. Observation drawn from multiple attempts.

**To Reproduce**
1. Download VHD to a storage account and create an Image.
2. Use the resource (image) in azhpc scripts to deploy a cluster

**Expected behavior**
1. Deployment should be successful

**Screenshots**
![image](https://user-images.githubusercontent.com/69217314/100675271-7f6e4a00-331b-11eb-8cae-63fbb19138d7.png)

**Configuration (please complete the following information):**
 - OS and version: CentOS 8.1 HPC Internal

souvik-de:
I don't think that this is specifically for the internal image but 8.1 in general

xpillons:
Thank you for raising this issue. As of today we don't support CentOS 8.1. In the meantime, you can copy the pbsserver.sh script in your own scripts directory and update it for 8.1. You may want to use PBS 20 as 19 doesn't support 8.1.

PR #467, opened 2021-02-11
First support of CentOS 8.1 and OpenPBS 20

xpillons:
Basic support of CentsOS 8.1 HPC Image in the **simple_hpc_pbs** example
- added pipeline test for CentOS 8.1 HPC Image
- added support of OpenPBS 20
- simplified PBS install scripts. New parameter introduced to specify the version (19 or 20) with 19 being the default
- added script dependency to pbsdownload.sh for pbsclient.sh and pbsserver.sh
- removed all copy operations from config files for PBS clusters
- removed all hardcoded RPM versions from config files for PBS clusters
- added new azhpc variable for PBS version in the **simple_hpc_pbs** example
- added scripts/azhpc-library.sh which contains help functions to retrieve os version and do a retry on commands
- pbsclient.sh and pbsserver.sh are using the retry function defined in azhpc-library.sh
- install-nfsserver.sh updated to support CentOS 7.x and CentOS 8.x

Closing #425 and #466

### DIFF
--- a/scripts/install-nfsserver.sh
+++ b/scripts/install-nfsserver.sh
@@ -72,14 +72,15 @@ exportfs
 ########################################
 # Tune NFS
 ########################################
-if is_centos7; then
-    cores=$(grep processor /proc/cpuinfo | wc -l)
-    nfs_proc=$(($cores * 4))
-    replace="s/#RPCNFSDCOUNT=16/RPCNFSDCOUNT=$nfs_proc/g"
-    sed -i -e "$replace" /etc/sysconfig/nfs
-    grep RPCNFSDCOUNT /etc/sysconfig/nfs
-fi
+
+cores=$(grep processor /proc/cpuinfo | wc -l)
+nfs_proc=$(($cores * 4))
+replace="s/# threads=8/threads=$nfs_proc/g"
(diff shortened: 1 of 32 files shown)


scripts/install-nfsserver.sh
```diff
@@ -54,11 +72,13 @@ exportfs
 ########################################
 # Tune NFS
 ########################################
-cores=$(grep processor /proc/cpuinfo | wc -l)
-nfs_proc=$(($cores * 4))
-replace="s/#RPCNFSDCOUNT=16/RPCNFSDCOUNT=$nfs_proc/g"
-sed -i -e "$replace" /etc/sysconfig/nfs
-grep RPCNFSDCOUNT /etc/sysconfig/nfs
+if is_centos7; then
```
edwardsp, line 75: you should still increase the number of threads in centos8.  This is done by setting "threads" in the "[nfsd]" section in /etc/nfs.conf.

edwardsp (changes requested):
looks good - although we should make sure we still increase the nfsd threads with centos 8

edwardsp (approved):
looks good

PR #478, opened 2021-03-18
Update run_ring_osu_bw_hpcx_pbs.sh

arstgr:
replace allreduce with bidirectional bandwidth test, add walltime limits to the current tests so that they don't continue forever, and report the possible errors into the screen and an error log file.


arstgr:
Hi @JonShelley Can you please review these changes?

xpillons (approved):

PR #481, opened 2021-03-26
Added more WRF example

xpillons:
- upgraded versions in spack builds
- fix pipeline
- added example with real life data

### DIFF
--- a/apps/spack/build_spack.sh
+++ b/apps/spack/build_spack.sh
@@ -1,9 +1,10 @@
 #!/bin/bash
+set -e
 APPS_SPACK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 APP_NAME=spack
 APP_VERSION=0.16.0
 SHARED_APP=${SHARED_APP:-/apps}
-INTEL_MPI_VERSION=${INTEL_MPI_VERSION:-2020.2.254}
+INTEL_MPI_VERSION=${INTEL_MPI_VERSION:-2018.4.274} # This should be retrieved dynamically
 
 MODULE_DIR=${SHARED_APP}/modulefiles/${APP_NAME}
 MODULE_NAME=${APP_NAME}_${APP_VERSION}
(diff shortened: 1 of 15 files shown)


apps/spack/build_spack.sh
```diff
@@ -1,9 +1,10 @@
 #!/bin/bash
+set -e
 APPS_SPACK_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 APP_NAME=spack
 APP_VERSION=0.16.0
 SHARED_APP=${SHARED_APP:-/apps}
-INTEL_MPI_VERSION=${INTEL_MPI_VERSION:-2020.2.254}
+INTEL_MPI_VERSION=${INTEL_MPI_VERSION:-2018.4.274} # This should be retrieved dynamically
```
garvct, line 7: Why change the Intel MPI library to an older version?

apps/wrf/example/namelist.tpl.input
```diff
@@ -0,0 +1,94 @@
+&time_control
```
garvct, line 1: Is this example input file useful? It alone cannot be used to run a job?

apps/wrf/example/namelist.tpl.wps
```diff
@@ -0,0 +1,39 @@
+&share
```
garvct, line 1: Is this example file useful? It alone cannot be used to generate a WRF input file?

garvct (commented):
Are all the WRF examples documented?

apps/spack/build_spack.sh
xpillons: because 2020 is no longer in the latest images !!!

apps/wrf/example/namelist.tpl.wps
xpillons: the namelist.wps file is used by the preparation tools and is needed to create the WRF input

xpillons:
example documenation is still WIP

garvct (approved):

PR #485, opened 2021-04-09
update build_wrf.sh

arstgr:
the changes allow the installation of jasper-devel libpng-devel on centOS8 (sudo dnf config-manager --set-enabled PowerTools is necessary)
the second change prevents ln from linking directories (which causes an error): ln -sf $(ls -p $NETCDF_C/lib/* | grep / ) $NETCDF/lib/


xpillons (approved):
good for me

PR #514, opened 2021-07-24
Create readme.md

mengruts:


mengruts:
Created "build_voltus.sh"

apps/cadence_voltus/build_voltus.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+# TODO: install directory
+INSTALL_DIR="/netapps/voltus/"
```
edwardsp, line 4: Is this correct?  It's not a default path that is created

mengruts: I made change to /voltus to make less confusion.

edwardsp: if you make this /apps/voltus it will be exported to all the nodes for the standard nfsserver setup

PR #551, opened 2021-11-29
Added get_physicalhostname to GPU_Monitoring

garvct:
- Get the physical host name and add it to the json that's pushed to Azure log analytics.


xpillons (approved):
Looks good for me. I hope the bash translation to python is correct
strings /var/lib/hyperv/.kvp_pool_3 | grep -A1 PhysicalHostName | head -n 2 | tail -1

PR #557, opened 2022-01-26
Cyclecloud slurm nhc (for NDv4(A100))

garvct:
Cyclecloud+SLURM cluster-init project for LBNL node health check framework.
-integrated with SLURM
-unhealthy nodes are put into a DRAIN state (when issue is resolved node will be moved back to IDLE state)
-Includes specific GPU cuda bandwidth test.


xpillons (approved):
how does nodes are set to DRAIN ?

PR #585, opened 2022-05-17
Add script to get physical host names. (to ndv4_a100_healthchecks)

garvct:
Added 
get_physicalhostname.py

cycleadmin@ip-0A2509BC:~$ ./get_physicalhostname.py
ip-0A2509BC physicalhostname = PHX801091908003


xpillons:
However I've been told that this is the cluster name not the hostname

PR #590, opened 2022-05-23
Added script to generate hpc diagnostics file

garvct:
Script will run pre-installed (HPC Marketplace images (e.g ubunt-hpc)) HPC Diagnostics (Useful when reporting an unhealthy node) and deposit the compressed tar file in directory.
Takes one argument the slurm hostname.

### DIFF
--- a/experimental/ndv4_a100_healthchecks/gen_hpcdiags.sh
+++ b/experimental/ndv4_a100_healthchecks/gen_hpcdiags.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 
 #Run pre-installed hpc diagnostics and deposit the compressed tar file in OUTDIR
+#
+# takes one argument, the slurm hostname.
 
 USER=cormac
 OUTDIR=/shared/home/${USER}/hpcdiags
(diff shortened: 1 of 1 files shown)


experimental/ndv4_a100_healthchecks/gen_hpcdiags.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
```
vanzod, line 2: I would recommend adding a `set -euo pipefail` here to stop the script any time an error occurs.

experimental/ndv4_a100_healthchecks/gen_hpcdiags.sh
```diff
@@ -0,0 +1,22 @@
...
+#Run pre-installed hpc diagnostics and deposit the compressed tar file in OUTDIR
+#
+# takes one argument, the slurm hostname.
+
+USER=cormac
+OUTDIR=/shared/home/${USER}/hpcdiags
```
vanzod, line 8: How about creating the output in the directory where the script is called from? This would allow to just execute the script without having to edit it beforehand.

PR #604, opened 2022-06-10
Added Support for a customized template file to create a cyclecloud cluster

garvct:
- Added support to create a cyclecloud cluster with a customized template file.

If the cyclecloud parameter "template" (in the config.json file) is set to the name of a cyclecloud text template file, it will use that to create the cyclecloud cluster (instead of a built-in cyclecloud template).

"template" : <"Name of the cyclecloud template file">

The format of the template file is <Scheduler_type>_<file_name>.txt
where scheduler can be "pbs" or "slurm". (e.g slurm_ndv4.txt)


xpillons (approved):
good for me

PR #608, opened 2022-06-11
Initial version of NDv4+cycle+anf+nhc deployment

garvct:
Complete E2E deployment of an NDv4 cluster leveraging Cyclecloud

The NDv4 cluster will consist of
- NDv4 compute nodes running Ubunt-hpc 18.04, Max GPU app clock freqencies are set, GPU persistent mode is enabled, local NVMe SSD's are mounted.
- Cyclecloud 8.2.2 with SLURM 2.6.4(pmix support) is installed and the pmix libraries are installed on NDv4 nodes.
- Automatic recovery from a reboot enabled (e.g NVMe SSD will be remounted and GPU clock freq reset)
- Premium SSD's used for OS disks, with larger capacity (60GB)
- Accelerated networking is enabled on NDv4
- User home directories are mounted on Azure netapp files
- Extensive automatic pre-job healthchecks are enabled (Unhealthy nodes will be put into a DRAIN state)
- A separate login node (separate from the scheduler is deployed)
- Cyclecloud autoscaling is disabled.
- Windows server (winbox) is deployed to access the Cyclecloud portal via RDP


xpillons (approved):
This is becoming a pretty complex deployment. Why not having this as well in azhop which could be useful for all?

PR #615, opened 2022-06-23
Fix NHC GPU application clocks script to iterate across all GPUs

vanzod:
The return statements in the `azure_gpu_app_clocks.nhc` script stop the loop at the first iteration, causing only GPU #0 to have the correct max application frequency to be correct.


garvct (approved):
Looks good.

PR #643, opened 2022-07-25
GPU Monitoring (Added option to collect IB metrics)

garvct:
- Added option to collect IB metrics (-ibm)


experimental/gpu_monitoring/gpu_data_collector.py
```diff
@@ -140,6 +153,55 @@ def get_physicalhostname():
...
 
 
+def get_counter_value(file_path):
+    file = open(file_path, "r")
+    value = file.read()
+    file.close()
+    return int(value)
+
+
+def get_counter_value(file_path):
```
vanzod, line 163: It looks like this function has been declared twice

garvct: I removed the duplicate "get_counter_value" function.

experimental/gpu_monitoring/gpu_data_collector.py
```diff
@@ -140,6 +153,48 @@ def get_physicalhostname():
...
+    file = open(file_path, "r")
+    value = file.read()
+    file.close()
+    return int(value)
+
+
+def infiniband_rate(current_counter, previous_counter, time_interval):
+    counter_delta = current_counter - previous_counter
+    if counter_delta < 0:
+        ib_counter_rate = 0
```
vanzod, line 166: I guess this is done because the counter resets when it reaches the bit-representable integer limit. 
Wouldn't it be better to set `ib_counter_rate = current_counter` instead of zero?
A more complete way would be to also add `limit - previous_counter`, but I do not know if we know the value of `limit`.

garvct: Changed to
if counter_delta < 0:
        ib_counter_rate = int((2*64 + counter_delta) / time_interval))

vanzod (approved):

PR #648, opened 2022-07-28
CC_SLURM_NHC (Added check for GPU Xid errors)

garvct:
- Added a test to check for GPU Xid errors.

### DIFF
--- a/experimental/cc_slurm_nhc/cc_slurm_nhc/specs/default/cluster-init/files/azure_gpu_xid.nhc
+++ b/experimental/cc_slurm_nhc/cc_slurm_nhc/specs/default/cluster-init/files/azure_gpu_xid.nhc
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Check for any GPU Xid errors in dmesg
+# Check for the following GPU Xid errors in dmesg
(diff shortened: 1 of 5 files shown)


vanzod:
A potential problem is that NHC will keep the node drained until logrotate archives syslog since `check_gpu_xid()` grep on the whole file. While this is desirable for uncorrectable errors requiring node OFR, for other less critical issues (e.g. user app error, driver GPU programming error, etc.) this could be unnecessary.

garvct:
I switched to using dmesg to check for Xid error codes and only Xid errors associated with H/W issues.

experimental/cc_slurm_nhc/cc_slurm_nhc/specs/default/cluster-init/files/azure_gpu_xid.nhc
```diff
@@ -1,20 +1,27 @@
 #!/bin/bash
 
-# Check for any GPU Xid errors in syslog
+# Check for any GPU Xid errors in dmesg
```
vanzod, line 3: "Check for the **following** GPU Xid errors in dmesg"

vanzod (approved):

PR #653, opened 2022-08-11
cc slurm nhc (Added single node NCCL all-reduce test)

garvct:
- Added single node NCCL all-reduce test
- Bug fix applied to drop cache test.
- Increased NHC timeout value
- Changed order of tests in the config file, run drop cache test first.


experimental/cc_slurm_nhc/cc_slurm_nhc/specs/default/cluster-init/files/azure_nccl_allreduce.nhc
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+  
+# Check for NVlink ssues by running NCCL allreduce.
```
vanzod, line 3: Since the file name is not too descriptive, I would add here that this runs on a single node through NVLink instead of IB

vanzod (approved):

PR #658, opened 2022-08-29
Add peering gateway opts

yosoyjay:
Expose additional peering configuration options to eliminate need to manually configure peering with existing vnets.
    
    - Add config dictionary in ['vnet']['peer'] for gateway options
    - Add config fields in ['vnet']['peer'][<peer-name>] for traffic options
    - Add method to delete peerings with azhpc-destroy when defined in config.json
    - Default configs remain unchanged
    - Update to README.md to reflect changes with example of adding peering

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,10 +94,26 @@ The config file will create or reuse vnet and subnets from the config file.
 
 This dictionary describes the virtual network peering to be created
 
-| Name               | Description                                                                        | Required | Default |
-|--------------------|------------------------------------------------------------------------------------|----------|---------|
-| **resource_group** | Name of the resource group containing the vnet to peer to                          |   yes    |         |
-| **vnet_name**      | Name of the vnet to peer to                                                        |   yes    |         |
+| Name                             | Description                                                          | Required | Default |
+|----------------------------------|----------------------------------------------------------------------|----------|---------|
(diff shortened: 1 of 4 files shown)


xpillons (approved):
sounds good to me, however i'm not able to test it

yosoyjay:
Good point that I meant to bring up with the PR, is there a preferred way to add tests?

xpillons:
> Good point that I meant to bring up with the PR, is there a preferred way to add tests?

Not really. We trust your tests

yosoyjay:
Okay, I'll add some testing.

yosoyjay:
How are tests run?  I see the `ci` dir, various bash scripts, what looks like CI yaml config files, but I'm not sure at all how this is actually run.

yosoyjay:
@xpillons I tested in a couple of different VPNs and it works as I expect.  However, an interesting side effect is that if you create a new VNet and create a peering to the another VNet that contains a VPN gateway with an active VPN connection, it drops the connection. You can immediately reconnect, but it is not ideal.  Do you have any idea about how to eliminate behavior?

xpillons:
I've seen this too and I don't know how to fix that. It's good to merge for me now.

yosoyjay:
Okay, thanks for confirming the behavior is not just limited to that.  I don't have permissions to merge, so you'll have to do that.

PR #662, opened 2022-09-08
Store private SSH key in Key Vault for bastion-jumpbox

vanzod:
AzHPC will create a new secret containing the local private SSH key to access the `bastion-jumpbox` VM.


scripts/add_secret_in_keyvault.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
garvct, line 1: scripts/create_password_in_keyvault.sh has very similar functionality, would it be better to modify this script and add an optional new_secret argument (so not be break existing scripts) instead of creating a new script?

garvct (approved):

PR #671, opened 2022-09-21
GPU Monitoring (Added CPU metrics)

garvct:
- Added cpu metrics (User, system, idle and iowait time, cpu loadavg)
- Added CPU memory (default:Total mem & free memory, many others can be monitored)

### DIFF
--- a/experimental/gpu_monitoring/gpu_data_collector.py
+++ b/experimental/gpu_monitoring/gpu_data_collector.py
@@ -58,6 +58,12 @@ ETH_COUNTERS = [
 #    'rx_errors',
 #    'rx_dropped',
 
+CPU_MEM_COUNTERS = [
+                'MemTotal',
+                'MemFree'
+              ]
+# Other useful CPU memory counters are Bufferes and Cached (there are many other useful counters see /proc/meminfo
+
 
 # Build the API signature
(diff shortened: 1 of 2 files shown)


experimental/gpu_monitoring/gpu_data_collector.py
```diff
@@ -303,12 +313,72 @@ def get_nfs_rates(nfs_counters, time_interval_seconds, hostname, physicalhostnam
...
+
+def get_cpu_loadavg_data():
+    cpu_loadavg_l = []
+    cpu_loadavg_l = read_file("/proc/loadavg")
+    value = cpu_loadavg_l[0].split()[0]
+
+    return value
+
+
+def get_cpu_data(hostname, physicalhostname_val, have_jobid, slurm_jobid):
```
vanzod, line 360: Thoughts about converting frequency to seconds?
The number of clock ticks per second can be obtained via `getconf CLK_TCK`.

garvct: I prefer to dump the raw data to log analytics (for speed) and do the conversion (whatever unit is of interest) in the log analytics kusto query. I think this conversion is fairly straightforward with USER_HZ = 1/100 sec.

PR #681, opened 2022-10-12
Add ndv4 slurm cc vpn

yosoyjay:
Add config and scripts I use to deploy NDv4 CycleCloud cluster using VPN.


yosoyjay:
@garvct A bit late, but here is the config I use for deploying clusters.  I don't have permissions on the repo, otherwise I'd assign you as this is derived from the `deploy_cycle_slurm_ndv4`.

garvct:
Can these modifications be added to the deploy_cycle_slurm_ndv4 instead of creating a new directory? It's the same deployment only using vpn instead of bastion?

yosoyjay:
It's substantially different.  Normally I'd keep in a different repo, but open this PR in response to the request to add it here.

PR #683, opened 2022-10-17
fairseq_moe_docker_slurm (Remove Slurm pinning)

garvct:
Removed Slurm pinning, we are using torch.distributed.launch with slurm (srun) starting 1 process per node. The slurm pinning is for 8 slurm tasks and so is unoptimal for a single slurm (srun) process.


yosoyjay:
Looks good to me.  @garvct did you test the impact on performance by chance?

JingchaoZhang (approved):

PR #688, opened 2022-10-19
added scripts to run fairseq with AML

JingchaoZhang:


experimental/fairseq_moe_docker_slurm/environment/context/Dockerfile
```diff
@@ -0,0 +1,20 @@
+ARG FROM_IMAGE_NAME=nvcr.io/nvidia/pytorch:21.10-py3
+  
+FROM ${FROM_IMAGE_NAME}
+
+ENV NCCL_DEBUG="WARN"
+ENV UCX_TLS="tcp"
+ENV UCX_NET_DEVICES="eth0"
+ENV NCCL_SOCKET_IFNAME="eth0"
```
yosoyjay, line 8: Is there an option to use IB here?

experimental/fairseq_moe_docker_slurm/environment/context/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ENV CUDA_DEVICE_ORDER="PCI_BUS_ID"
+ENV NCCL_IB_PCI_RELAXED_ORDERING=1
+ENV NCCL_TOPO_FILE="/opt/microsoft/ndv4-topo.xml"
+
+RUN apt-get -y install build-essential
+RUN pip install fairscale==0.4.0
+RUN pip install hydra-core==1.0.7 omegaconf==2.0.6
+RUN pip install boto3
```
yosoyjay, line 16: Combining these libs into a single line, if possible, will reduce the number of layers in the image which is generally a positive thing.  Might also be a good idea to instead put in a requirements.txt and build from that.

experimental/fairseq_moe_docker_slurm/run_fairseq_moe_aml.py
```diff
@@ -0,0 +1,136 @@
...
+    compute_target=compute_target,
+    environment=fairseq_env,
+    distributed_job_config=distr_config
+    )
+
+#-------------------------------------------------------------------------------
+# Submit experiment
+#-------------------------------------------------------------------------------
+# Name your experiment by setting YOUR_EXPERIMENT_NAME
+run = Experiment(ws, 'YOUR_EXPERIMENT_NAME').submit(fairseq_src)
```
yosoyjay, line 135: I'd suggest adding CLI args (e.g. from `argparse` or `click`) that can be passed into a main function instead of having to edit this script for different experiments.

yosoyjay (commented):
Looks pretty good.  I made some small comments on Dockerfile composition that would make some small improvements.  I do think the `run_fairseq_moe_aml.py` driving script would be improved by adding CLI args to eliminate hard-coded assumptions.

experimental/fairseq_moe_docker_slurm/environment/context/Dockerfile
JingchaoZhang: I thought about the same thing. These variable settings are inherited from the original setup. My guess is that these communications do not involve RDMA so diverted to ethernet port. @garvct can comment on it.

experimental/fairseq_moe_docker_slurm/environment/context/Dockerfile
JingchaoZhang: Good point. Thanks!

experimental/fairseq_moe_docker_slurm/run_fairseq_moe_aml.py
JingchaoZhang: Good point. Thanks!

yosoyjay (approved):
Looks good

experimental/fairseq_moe_docker_slurm/environment/context/Dockerfile
yosoyjay: Yeah, I've seen it elsewhere and I was immediately curious about it.  Even without RDMA, IB will have high bandwidth and lower latency than using eth0.

PR #693, opened 2022-11-14
Nsg multiple sourceip and rocky image plan

hmeiland:
update api version and change prefix to prefixes for nsg
and add rocky plan

### DIFF
--- a/pyazhpc/arm.py
+++ b/pyazhpc/arm.py
@@ -1007,7 +1007,7 @@ class ArmTemplate:
 
         # Add support for cyclecloud/azurehpc plan
         plan = ""
-        if ros[0] == "azurecyclecloud" or ros[0] == "azhpc" or ros[0] == "almalinux":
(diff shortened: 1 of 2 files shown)


pyazhpc/arm.py
```diff
@@ -689,7 +689,7 @@ def _add_vm(self, cfg, r, vnet_in_deployment):
                             "protocol": "Tcp",
                             "sourcePortRange": "*",
                             "destinationPortRange": "3389",
-                            "sourceAddressPrefix": "*",
+                            "sourceAddressPrefixes": "*",
```
edwardsp, line 692: This is type "string[]" and so needs a list of strings.

edwardsp (commented):
I'm not sure how you are using this as there is no example with multiple source address prefixes - how are they being assigned?  Also, throughout you are assigning a string value, which is valid for `sourceAddressPrefix`, but `sourceAddressPrefixes` takes a string list.

pyazhpc/arm.py
hmeiland: updated

hmeiland:
adding vnc nsg definition

edwardsp (approved):
looks good

PR #707, opened 2023-02-07
Added RAID health check test

vanzod:
Add test to check RAID array and corresponding file system health.
Performed tests:

1. Check that the file system corresponding to the provided mount point is a RAID device
2. Check that mdadm returns clean array state
3. Check that the array contains the specified number of devices (disks)
4. Check that file can be written at mount point
5. Check that file can be deletes at mount point


yosoyjay (approved):
lgtm

garvct (approved):

PR #735, opened 2023-11-21
Aks ndv4 update

garvct:
- Remove mofed installation script.
- Removed hpc-diagnostics.sh script.
- Removed custom-test-setup.sh script.
- Corrected ib_write_bw_gdr.nhc script
- Updated Dockerfile


JingchaoZhang:
Two optional updates:
1. This commit is for `ndv4`, but it's worth pointing out the current setup works for `nd96asr_v4` specifically. If one uses `nd96amsr_a100_v4`, then the `conf/nd96amsr_a100_v4.conf` file needs to be updated before the NHC can run. 
2. It's probably a good idea to add a `sleep 3600` in the `hpc-diagnostics.yaml` file. This way people can `kubectl cp` the tarball directly after the log collection.

PR #736, opened 2023-11-21
aks_ndv5 (Run NHC and hpc-diagnostics on AKS+NDv5)

garvct:
Dockerfile, scripts and yaml manifects to run NHC healthchecks and hpc-diagnostics on AKS+NDv5


JingchaoZhang (commented):
need to copy `[nhc.yaml](https://github.com/Azure/azurehpc/blob/aks_ndv5/experimental/aks_ndv4/nhc.yaml)` from aks_ndv4 as well.

JingchaoZhang (approved):

PR #737, opened 2023-12-19
Update Pyxis/Enroot CC project

vanzod:
- Update to Pyxis 0.16.1 and Enroot 3.4.1
- Ensure compatibility with CycleCloud v8.4+ and Ubuntu-HPC 22.4 images


yosoyjay (commented):
This works.  We should maybe look at testing for scheduler by testing for presence of specific files via `test`, but effectively the same as `ls` and `grep`.

xpillons:
does it means that Pyxis and enroot are not in the hpc images ?

PR #738, opened 2023-12-21
Install pyxis and enroot from local packages

vanzod:
Make Pyxis and Enroot installation more robust against network, DNS or GitHub issues by using local packages instead of downloading them.


yosoyjay (approved):
Looks good.

